PR #73, opened 2010-09-06
Fix locales for rails 2.3.9

jonnii:
The i18n stuff has changed for rails 2.3.9, it's using the new %{} syntax over the {{}} syntax like before.  This fixes the default en.yml to reflect that.


carlosantoniodasilva:
Applied in 900e915851b5ac994e34c15e7839deb4b500ad56, thanks.

jonnii:
No problem

PR #86, opened 2010-10-19
support for shallowed urls

dmitry-a-l:
this patch adds support for shallowed urls 
enable shallowed url in controller by passing :shallow => true on belong_to macro


dmitry-a-l:
denied due new version

PR #93, opened 2010-11-12
Redirect to index action if show action not defined

janisr:


josevalim:
Thanks for the pull request. Could you please include some tests as well?

PR #107, opened 2011-01-03
replace deprecated class_inheritable_* methods with class_attribute

etehtsea:
May be it's not a good idea to undefine query_methods, but tests are falling without.


josevalim:
Thanks, applied.

PR #112, opened 2011-01-14
Better support for namespaced controllers

sirupsen:
I had trouble with namespaced controllers not picking up the request params. I achieved the wished behavior by changing the instance name, however to me it seems odd that the request params fetching relies on something freely changeable by the user. In my patch the request params are extracted directly from the controller name, now users can change the instance name as they like without breaking Inherited Resources.

Furthermore, I included a model prioritizing system for namespaced controllers. If you have a namespaced Forum posts controller like this:

```
class Forum::PostsController < InheritedResources::Base
end
```

Inherited Resources with this patch prioritizes the resource class for the namespaced controller in
this order:

```
Forum::Post 
ForumPost   
Post
```

In lieu of going directly for the User model as current `master` does.

This should close issues like these:
https://github.com/josevalim/inherited_resources/issues/issue/64
Should now work as the issue author described, out the box.

https://github.com/josevalim/inherited_resources/issues#issue/109
Instance name issue

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -463,6 +463,20 @@ option.
 
 It will deal with everything again and hide the action :index from you.
 
+== Namespaced Controllers
+
+Namespaced controllers works out the box.
+
+    class Forum::PostsController < InheritedResources::Base
+    end
+
+Inherited Resources prioritizes the default resource class for the namespaced controller in
+this order:
+
+    Forum::Post
+    ForumPost
+    Post
+
 == URL Helpers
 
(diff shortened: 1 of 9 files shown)


iain:
+1, although I don't understand why you've put `AdminUser` in here. It looks out of place.

sirupsen:
It's there because some people do not like namespaced models. Although the example could be better, e.g. Forum::Post, ForumPost and Post makes more sense. I have updated with that, better, example.

josevalim:
Awesome, merged!

PR #113, opened 2011-01-21
Custom actins, smart redirects

dmitry-a-l:
# Custom actions

Now we can define custom actions in controller:

```
class ButtonsController < InheritedResources::Base
  custom_actions :resource => :delete, :collection => :search
end
```

This code creates `delete` and `search` actions in controller (they behaves like `show` and `index` accordingly). Also, it will produce `delete_resource_{path,url}` and `search_resources_{path,url}` url helpers.
# Smart redirects

Redirects in actions calcultes depending on the existent controller methods. 
Redirects in `create` and `update` actions calculates in following order `resource_url`, `collection_url`, `parent_url`, `root_url` . Redirect in `destroy` action calculate in following order `collection_url`, `parent_url`, `root_url`.

Example

```
ButtonsConntroller  < InheritedResources::Base
  belong_to :window
  actions :all, :except => [:show, :index]
end
```

This controller redirect to parent window after all CUD actions


josevalim:
Awesome, applied!

jackdempsey:
Ida,

Thanks for this great patch! I'm a bit confused how that example shows off the power of smart redirects. 

I've gotten this working locally when I define a collection_url method that just users the resources _url method. 

How does actions:all, :except => [:show, :index] demonstrate this? Sorry if I'm being obtuse, but I don't quite see it. Any tips?

Thanks!

jackdempsey:
Looking at this a bit more, it seems that you can use actions :all, :except => [:show] to undef show...and smart_resource_url is smart enough to say "hey, they don't have a show, how about we try the collection url"

But what if you want that show action? You just don't want to redirect to it after you create. 

Seems you either override resource_url in that controller to point to collection_url or you need to do something like the CollectionResponder?

Ah, and because of this line

options[:location] ||= smart_resource_url

collection responder will already see a location set and not do it's thing....

So in general because of this useful feature, if we want to redirect to the collection we have to override resource_url each time, huh. Not that it's hard, you can easily alias_method :resource_url, :collection_url or something, but I just want to make sure I understand this.

dmitry-a-l:
Hi jackdempsey.

This patch is a logical continuation of [sviesczs idea](https://github.com/josevalim/inherited_resources/commit/e11b02e63e83cf4ccdcbee4799868c1ec02906e3)
The main goal of this patch is reducing amount of code in following common cases:
- You have not `show` action. So after `create` (or `update`) you must be redirected to `index`
- You have not `show` and `index` actions, but you have parent controller. So after `create` you must be redirected to parent_url. 
- You have not `show` and `index` actions, and you haven't parent controller. So after `create` you must be redirected to root_url.
- Any combination of absent actions.

Thus if we don't know what actions we have in runtime, we must try all of this.  So I introduce two "smart" methods: `smart_resource_url` and `smart_collection_url` applied only in `create`, `update`, and `destroy` actions accordingly.

I strongly don't recommendate to alias `resource_url` and `collection_url` because you will have complicated views and unpredictable errors after doing this. So if you don't want to redirect to show after create just use this code 

```
def create
  create! { collection_url }
end
```

jackdempsey:
Hi Ida,

Thanks for the response, I appreciate it. 

That all makes good sense. The only piece I don't like is the end solution. I don't want to have to manually define create and update methods in each controller each time I want to redirect to collection_url.

I don't like aliasing the resource_url to collection_url of course, but think there has to be the possibility for a good solution in between the two. 

Do you have any thoughts on what a good system level solution would look like? Something like the collection responder ability?

dmitry-a-l:
Ok, the apprecable solution might be inheritance (or modules). For example:

```
ApplicationController < InheritedResources::Base
  def create
    create! { collection_url }
  end
end

MyCustomController < ApplicationController
end
```

PS: dude, my nick is downcased LDA (not IDA)  :)

dmitry-a-l:
Or you can introduce some DSL to IR. I think it would be great if we will have nice DSL syntax like 

```
MyCustomController < InheritedResources::Base
  redirects_to :index, :after => [:create, :update]
end
```

I'm too busy now to implement this feature

jackdempsey:
Ida,

Hmm, I'll have to think about that first idea. I think you might have just triggered a good idea on some other refactorings.

I like the second one too though. I was thinking about DSL just like that. I'll see if I can put together something for you and José to take a peek at. 

Thanks!

PR #116, opened 2011-01-25
Use "scoped" instead of "all" method for collection.

amw:
Currently default index action executes the load query inside it's body instead of where the collection rows are first used - inside the view. Model.all or Model.-finder-.all cause the default or current scope to be transformed into an Array object with result's rows.

Instead one can use Model.scoped or Model.-finder-.scoped to return the collection as ActiveRecord::Relation object. This is the same object that is returned by finders like Model.where().limit().offset() etc.. This way the rows won't be fetched until later. There are two benefits of this.

First this will be particularly important if automatic flushing gets implemented and released with Rails 3.1 (I don't know the status of that feature, but Katz wrote about it once). By postponing query execution we allow the layout's header to be sent to user's browser earlier.

Second you get a chance to add additional finders inside the view. I.e:
    <% @accounts.limit(5).each do |account| %>

Currently waiting for suggestions regarding test cases.


amw:
I have no idea why there are so many commits in this pull request. When I look on your master branch on github the last commit is dated 29th December, but my fork has 18 commits more, some of them very recent. And I've pulled it directly from your master branch. What's wrong?

josevalim:
There is no problem, I would still be able to apply your commit. But you probably merged my branch into yours, instead of calling rebase. If you want to start from scratch again, this shuold be your work flow:

  1) Fork on Github
  2) git clone fork@github
  3) git checkout -b scoped
  4) Do your changes
  5) git checkout master
  6) git remote add josevalim https://josevalim@github.com/josevalim/inherited_resources.git
  7) git pull josevalim master
  8) git checkout scoped
  9) git rebase master
  10) Push your branch and send a pull request

About test cases, they all mock Active Record, so if you run tests now, they will probably fail. You just need to fix them and make them pass.

josevalim:
Hey, I may do another release soon.  So if you can work out the tests (basically by fixing them) it is likely I can include your patches in the next release! thanks!

amw:
I'll try to do this today. Cheers

amw:
This latest commit fixes the existing tests, but it doesn't introduce a new one to catch future commits that could break the scoped collection return value.

Also for simplicity I made the :scoped expected return value an array, like the one returned from :all. I think that's ok since ActiveRecord::Relation accepts all the methods that Array responds to.

As to the commits count on this pull request - if you add my fork to your remotes you'll see that I haven't done any merges. It's a topic branch on top of your master. It looks like a github issue. Anyway - you shouldn't have any problems pulling to your local repository.

josevalim:
Sounds perfect, merged! Thanks!

brodock:
This fix introduced a regression: InheritedResources with ActiveResources returns now: undefined method `scoped' for _ModelName_:Class

amw:
Forgive my clueless comment before. I didn't expect inheritedresources to be used with activeresource. I'll check the rails code for activeresource and see how to play nice with it. But maybe we will need to fall back to respond_to? and use "scoped" or "all" accordingly.

amw:
I've added a new pull request to work around this. It still needs some work done with tests, but you can track the progress there:

https://github.com/josevalim/inherited_resources/pull/127

brodock:
Thanks... just to clarify, as you asked me, I'm using with rails 3.0.4

ianha:
I'm noticing i'm getting the error "undefined method 'scoped'" when trying to call UserSessionsContorller#index when using Authlogic.  Everything works fine for my other controllers.

Is there something Inherited Resources needs that Authlogic for some reason is overriding or not exposing?

I know it seems weird to index and view all your sessions, but I'm just hacking away to learn things and discovered this.

Thanks in advance!

amw:
There's a fix for this in master, but it wasn't released to rubygems yet. You can try using the gem directly from github and I'll ask Jose to release a new version.

amw:
@ianha Jose released a new gem version. It should contain the fix you're looking for.

Issue #95, opened 2010-11-16
Helpers are not refreshing in development mode


railsnerd:
Hi there

Normal development Rails behavior is for changes to a helper to be automatically reflected on page refresh. A server reload is not necessary.

However this gem needs a server reload for the helper to get updated.

Is there a work around for this?  I hate tested in Rails 3.1

thanks

railsnerd:
OK so I found this ticket https://github.com/josevalim/inherited_resources/issues/closed#issue/81

Is inheriting from inside the controller the official solution?  In the comments this is proposed as a temporary fix?

Would be nice to keep using "class X < InheritedResources::Base"

josevalim:
This is a bug and a patch is welcome.

collin:
I'm not sure how to fix this bug, but maybe this information helps:

Upon booting a rails console ( with X < InheritedResources::Base )

```
  > ActiveSupport::Dependencies.autoloaded_constants
 => ["Ability", "Article", "Publication", "Staff", "User"] 
  > StaffsController
 => StaffsController 
  > ActiveSupport::Dependencies.autoloaded_constants
 => ["Ability", "Article", "Publication", "Staff", "User", "DeviseHelper", "LinkHelper", "ProfileHelper", "ApplicationController", "StaffsController"] 
  > reload!
Reloading...
 => true 
  > StaffsController
 => StaffsController 
  > ActiveSupport::Dependencies.autoloaded_constants
 => ["Ability", "Article", "Publication", "Staff", "User", "StaffsController"]
```

I actually had my first debugging session that involved breaking of constants autoloading. In that case I was using require in a file that was being autoloaded, which short circuited the code in ActiveSupport::Dependencies that would have registered the class for reloading.

The fix I employed there was to use Kernel.require instead of require, which implicitly is called on Object. Rails overrides Object.require, and leaves Kernel.require alone.

I peeked into inherited_resources/base, but that strategy had no effect.

If somebody knows a good resource for which actions will short-circuit ActiveSupport::Dependencies that would be a great thing to have when approaching this bug.

albertoperdomo:
Is this still an issue? I'm seeing this behaviour as well...

borisnadion:
looks like it is

marcandre:
+1

escobera:
Need this too.

vitaly:
hey, just wasted lots of time debugging this, finally traced to inherited_resource. Any ETA on the fix?

[...]

PR #118, opened 2011-01-28
Fix for refreshing ApplicationHelper in development mode

fschwahn:
This is a fix for issue #95.
It essentially overwrites this method from ActionController::Base

```
def self.inherited(klass)
  super
  klass.helper :all if klass.superclass == ActionController::Base
end
```

I'm not sure if this is the most elegant fix, but it does work for me.

Best regards,
Fabian

### DIFF
--- a/lib/inherited_resources/base.rb
+++ b/lib/inherited_resources/base.rb
@@ -39,6 +39,11 @@ module InheritedResources
           :resource_class?, :parents_symbols?, :resources_configuration?
       end
     end
+    
+    def self.inherited(klass)
+      super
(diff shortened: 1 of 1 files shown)


josevalim:
Interesting to know that it solves the issue, but I am afraid it may cause side-effects, I will play with it later! Thanks a lot!

fschwahn:
Yes, I suspected that it could have side-effects. If it helps you fix the problem properly I'm satisfied!

josevalim:
Applied, thanks

PR #121, opened 2011-01-31
:param for :finder to base on

farnoy:
I've needed to find a model with `find_by_name` param, but it was still searching by `params[:id]`, so I patched in the `resource` method to use param from new function `param_name_for_finder`. It can be used like this:

`defaults :finder => :find_by_name, :param => :object_name`

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -41,13 +41,6 @@ module InheritedResources
         get_resource_ivar || set_resource_ivar(end_of_association_chain.send(method_for_find, params[param_name_for_find]))
       end
       
-      # Supplies finder with param name to search with.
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks for your pull request, but this is intentional. I am even surprised we have :finder as option. I usually advise people to simply overwrite resource, it is simpler and much clearer in intention.

farnoy:
I know conventions, but I don't agree with that. Sometimes is just cleaner to use `params[:object_name]` than using `params[:id]`, which is a string with name anyway, because you want to provide sweet url's. I understand your desicion, but that defaults to `:id`, so it wouldn't collide with anything.

josevalim:
You didn't understand me. I am not saying that it is wrong to use :object_name. I am just saying it is easier and better to overwrite the method resource than changing the key through an option.

PR #127, opened 2011-02-21
Scoped collection

amw:
User brodock has reported that my last patch doesn't work with ActiveResource which does not implement `scoped`. This commit makes us use `scoped` only when `end_of_association_chain` responds to it and `all` otherwise.

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -20,7 +20,10 @@ module InheritedResources
       #   end
       #
       def collection
-        get_collection_ivar || set_collection_ivar(end_of_association_chain.scoped)
+        get_collection_ivar || begin
+          c = end_of_association_chain
+          set_collection_ivar(c.respond_to?(:scoped) ? c.scoped : c.all)
+        end
       end
 
       # This is how the resource is loaded.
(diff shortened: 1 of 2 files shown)


josevalim:
Could you please add a test?

amw:
I will. I also have to investigate why the current tests produce errors. I've checked them only after sending the pull request.

josevalim:
They use mocks, so  it is probably choking on the respond_to? call. You just need to mock the appropriate things.

carlosantoniodasilva:
@amw This might be because of the operator precedence: && executes first than the assignment:

```
>> c = 'carlos' && c.upcase
NoMethodError: undefined method `upcase' for nil:NilClass
    from (irb):9
>> c = 'carlos' and c.upcase
=> "CARLOS"
```

josevalim:
Good call Carlos! Although I wouldn't use end, try using a begin/end block:

```
get_collection_ivar || begin
  collection = end_of_association_chain
  set_collection_ivar(collection.respond_to? :scoped ? collection.scoped : collection.all)
end
```

carlosantoniodasilva:
Ah sure, I didn't intend to say to use `and`, it was just to better explain the operator thing :).

amw:
You're definitely right. Thanks! I'll commit in a sec.

amw:
Fixed the commit and added the test.

PR #128, opened 2011-02-21
Some fixes to tests

amw:
If you'd use ruby 1.9.2 with inherited resources you'd get some errors and failures when running tests. Details available in commit messages.


josevalim:
Awesome, thanks!

PR #145, opened 2011-05-22
Turn InheritedResources::Base into a reloadable constant.

### LINKED ISSUE #144: Inherited controllers not inheriting ApplicationController methods
I'm sure I'm doing something wrong here but I noticed similar issues with InheritedResources controllers not being able to properly use ApplicationController methods or blocks. I'm still pretty new to Rails but here it goes.

Specifically, I'm implementing CanCan and trying to use rescue_from for AccessDenied exceptions. If I put the rescue_from block in ApplicationController, it is ignored. The exact same block works if I put it instead in the controller with load_and_authorize_resource.

Since InheritedResources::Base inherits from ApplicationController, it should be able to work with the rescue_block from ApplicationController, right?

thedarkone:
See discussion in #95, would also fix #144.

### DIFF
--- a/app/controllers/inherited_resources/base.rb
+++ b/app/controllers/inherited_resources/base.rb
@@ -1,6 +1,3 @@
-require 'inherited_resources/blank_slate'
-require 'inherited_resources/responder'
-
 module InheritedResources
   # = Base
   #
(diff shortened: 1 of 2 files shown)


jbbarth:
I confirm it works. Any plan to include it in a new release soon ?

PR #153, opened 2011-07-08
Added support for rails 3.1 new mass assignment conventions

somebody32:
I've added new method 'with_role' that can be used to set the role that will be used for 'new' and 'update_attrbutes' methods to satisfy new rails 3.1 mass assignment conventions

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -49,7 +49,7 @@ module InheritedResources
       # instance variable.
       #
       def build_resource
-        get_resource_ivar || set_resource_ivar(end_of_association_chain.send(method_for_build, resource_params))
+        get_resource_ivar || set_resource_ivar(end_of_association_chain.send(method_for_build, *resource_params))
       end
 
       # Responsible for saving the resource on :create method. Overwriting this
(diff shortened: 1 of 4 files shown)


josevalim:
perfect pull request, thanks. merged!

please ping me when Rails 3.1 is out to release a new version?

somebody32:
ping, rails 3.1 is out

josevalim:
Done! thanks for reminding me!

njakobsen:
How are you supposed to use this? I have multiple users with different roles using the same controller to update models. Some attributes are protected and for admins only.

I figured I'd make a before filter that runs the following

```
def set_role
  with_role(current_user.role)
end
```

But it looks like with_role is a class method you call on the controller. How do you apply this in a real use case?

Zequez:
Awesome pull! I was about to do it myself until I saw the changelog! ^^
Thanks!

jweir:
@njakobsen and anyone else trying to dynamically set the role

you might need to overwrite two instance methods:

```
 def role_given?;end # true or false
 def as_role?;end # returns a hash {:as => ROLE}
```

An example with only an :admin role supported

```
def role_given?
  user.is_admin? 
end

def as_role
  user.is_admin? ? {:as => :admin} : {}
end
```

PR #154, opened 2011-07-26
Allow to set roles dynamically

chipiga:
Now you can set with_role based on current user. Like this:

  with_role lambda{|c| c.current_user.try(:role)}
  with_role lambda{|c| c.current_user.try(:admin?) ? 'admin' : nil}


josevalim:
Why not simply override the as_role method?

Zequez:
Note that if you override the `as_role` method, you must override `role_given?` to return true.

PR #160, opened 2011-08-17
Build nested objects

dmitry-a-l:
This feature adds ability to build nested objects (defined with accepts_nested_attributes_for macro) in  appropriate controller methods


josevalim:
Thanks for the pull request, but this is actually a view concern. For example, you don't want your JSON/XML response  to have three empty resources.

jcampb:
maybe a "noob" question Jose, but when you say this is a view concern, I assume you mean that we should create the nested object within the view rather than the controller. 

Would you mind elaborating on how that might look? 
I'm trying:
    ...
    f.simple_fields_for Profile.new, :as=> :profile_attributes  do
    ...
but I get a AssociationTypeMismatch in the create action with ActiveSupport::HashWithIndifferentAccess

kuraga:
@joelmoss @lda I think this feature is very useful! Isn't it implemented yet?..

I have `comment`s and `attachment`s in my `article`. So I need to write

``` ruby
class ArticlesController < InheritedResources::Base
  def new
    @article = Article.new(params[:project])
    @article.build_comment
    5.times { @article.attachments.build }
    new!
  end
end
```

(where `5` is a constant parameter)

It's not good... Does this PR (by the way it's broken) `build` `accept_nested_attributes_for`'s nested objects? What about  `has_one` and `has_many`?

dmitry-a-l:
Hi, @kuraga. @josevalim is right, what do you send in JSON/XML response for articles#new ? PR broken because it is unsupported by me.

dmitry-a-l:
So, I think the best place for such things is a presenter layer.

kuraga:
@lda Sorry, but I don't understand. I use these variables to render a form (HTML).

Is presentation layer a view? Don't we `build` associations for `fields_for` in a controller?

I don't know where did I learn it. It was long time ago :) But regarding to [ASCIIcasts #196](http://asciicasts.com/episodes/196):

``` ruby
    def new  
      @survey = Survey.new  
      3.times { @survey.questions.build }  
    end
```

dmitry-a-l:
I think about using presenters, but I don't use its yet. Read more about presenter/decorator pattern in Google :) 

Sure, you CAN do this in controller, but probably (as @josevalim points) this is wrong place.

kuraga:
@lda @josevalim But many people (and guides' authors) do this. Is this feature as such as the whole thing that `inherited resources` does - to make controllers shortly, isn't it?

I think this feature may be optional... Assume `build_nested` macro...

masterdezign:
@kuraga do you mean that this feature is needed to avoid just this line?

```
3.times { @survey.questions.build }  
```

However, one may consider the following code in survey.rb model

```
def self.build
  survey = self.new
  3.times { survey.questions.build }  
  survey
end
```

Now we just have in survey_controller.rb

```
def new  
  @survey = Survey.build
  new!
end
```

How about that? I'm not quite sure inherited_resources should handle models like this.

kuraga:
@masterdezign but without this line I don't write `new` method at all.

> I'm not quite sure inherited_resources should handle models like this.

Why?

masterdezign:
@kuraga If one chooses any non-trivial model, this model is hard to be automated unless you come across a pattern. 
For instance, Rails automated many dev tasks, however, inherited_resources provides even more automation since CRUD patterns are defined.

The question is, should we extend functionality, write something on top of software we already have (like inherited_resources on top of Rails), or return once again to OOP principles when we are not able to solve the problem in batch?

richpeck:
As someone with this issue currently, I would recommend the OP's original request was to give a DRY way of handling the accepts_nested_attributes_for function. I'm building a "resources_controller.rb" controller - and it would be great if we just had to put "build_nested_objects" based on whether the relevant models had the accepts_nested_attributes_for function defined

PR #158, opened 2011-08-23
can't get the parent in a polymorphic_belongs_to 

indrekj:
I can't access the parent object when I use a polymorphic relation between my controllers.

Following the documentation, I have a controller like this:

``` ruby
class PhotosController < InheritedResources::Base
  belongs_to :user, :task, :polymorphic => true

  def index
    logger.debug(parent.inspect)
  end
end
```

But parent always returns nil. It doesn't work either with:

``` ruby
polymorphic_belongs_to :user, :task 
```

However, it works if I don't use a polymorphic relation:

``` ruby
belongs_to :user 
```

or

``` ruby
belongs_to :task 
```

Note that my routes and everything else regarding inherited_resources work fine.

I don't know if it is a bug, or if I misinterpreted  the documentation.

thanks for your help

julien

### DIFF
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -181,6 +181,7 @@ module InheritedResources
             nil
           end
 
+
           config[:collection_name] = options.delete(:collection_name) || symbol.to_s.pluralize.to_sym
           config[:instance_name]   = options.delete(:instance_name) || symbol
           config[:param]           = options.delete(:param) || :"#{symbol}_id"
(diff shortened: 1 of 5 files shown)


indrekj:
I think I found the problem.

``` ruby
# Here you can use parent in the views
def show
  show!
end
```

``` ruby
def show
  resource
  # now you can use parent too
end
```

``` ruby
def show
  parent # doesn't work, because it is not loaded. @parent_type is nil. It gets loaded after `resource` and after `show!`
end
```

I added a patch, but its code isn't very nice. Maybe you can get something out of it :).

jujudellago:
I tried to apply the patch and now it works as expected !
Thank you so so so much !!! 

(I closed the pull request by mistake, I don't think I should be the one doing that ?  anyway the pathe solves my problem...)

josevalim:
Could someone please rebase this patch? It would be a nice addition.

indrekj:
rebased: https://github.com/josevalim/inherited_resources/pull/189

PR #168, opened 2011-09-09
Namespaced models and controllers in a polymorphic belongs_to relationship don't work

gaizka:
In a polymorphic belongs_to relationship, you cannot specity parent_class
option. So, if the models are namespaces, you would not be able to find them.

When looking for the model, try the same namespace as the one the controller is.

### DIFF
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -178,7 +178,16 @@ module InheritedResources
             class_name.constantize
           rescue NameError => e
             raise unless e.message.include?(class_name)
-            nil
+
+            # Let's try to namespace the model with the controller's namespace
+            if self.name.match(/::/)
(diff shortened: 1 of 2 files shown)


ep-wac:
with a class Admin::OxenController and an AttachmentsController belongs_to :ox, :page, :optional => true, and calling `collection_url` on a /admin/oxen/1/attachments/new request, I'm left with

```
undefined method `ox_attachments_url' for #<AttachmentsController:0x007fd30095f120>
```

@gaizka I tried this commit but I probably does not understand it 'enough' - it still throws the above error

boie0025:
:+1:

ka8725:
:+1:

ka8725:
This code looks complicated because of nested rescue and doesn't contain the tests for changes. I've created other [pull request](https://github.com/josevalim/inherited_resources/pull/303) which resolves this problem. I hope it will be merged sometime.

rafaelfranca:
Closing because #303 was already merged.

PR #172, opened 2011-10-08
Make parent? work in a before filter for optional belongs_to associations.

jemmyw:
If a before filter wants to know if the resource is being requested with
a parent on an optional belongs_to association it should not have to
load the resource first. Previously parent? relied on loaded information
only.


jemmyw:
Actually this seems to be the same as #158

PR #191, opened 2011-12-27
Fix url helpers on mountable engines

juniorz:
Original pull request: https://github.com/josevalim/inherited_resources/pull/186 by @marcelloma

Changed to trust on a public API instead a obscure method (**_railtie**) and verifies if the engine has an isolated namespace.

### DIFF
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -338,6 +338,16 @@ module InheritedResources
 
         # Deal with namespaced controllers
         namespaces = self.controller_path.split('/')[0..-2]
+
+        # Remove namespace if its a mountable engine
+        namespaces.delete_if do |namespace|
+          begin
+            "#{namespace}/engine".camelize.constantize.isolated?
+          rescue
+            false
+          end
+        end
+
         config[:route_prefix] = namespaces.join('_') unless namespaces.empty?
 
         # Deal with default request parameters in namespaced controllers, e.g.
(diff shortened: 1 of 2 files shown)


lib/inherited_resources/class_methods.rb
```diff
@@ -338,6 +338,16 @@ def initialize_resources_class_accessors! #:nodoc:
 
         # Deal with namespaced controllers
         namespaces = self.controller_path.split('/')[0..-2]
+
+        # Remove namespace if its a mountable engine
+        namespaces.delete_if do |namespace|
```
josevalim, line 343: I believe the best way to achieve this is to simply do it as Rails:

https://github.com/rails/rails/blob/master/activemodel/lib/active_model/naming.rb#L91

Something like this:

```
namespaces = self.parents
self.parents.delete_if do |n|
    n.respond_to?(:use_relative_model_naming?) && n.use_relative_model_naming?
end
namespaces.map!(&:underscore)
```

It will work only on 3.2 though but I believe this is fine.

knewter:
or +1 on this one :)

prikha:
Please post any temporary solution for this unless there no way out...

juniorz:
That is what I made: https://github.com/juniorz/inherited_resources/commit/9b51bc3cf588cd07a67b0dc719fcd075efcd16ef

But, it looks like this gem is "deprecated" and will no longer be updated...

panyamin:
+1 this fix is very needed.

_Works on mountable engine (rails 4.0.0)_

pusewicz:
:+1:

rafaelfranca:
This need a rebase.

PR #201, opened 2012-01-30
pass all actions as argument instead of relying on the prepended helper....

jachenry:
...  prepending the action to _polymorphic_path works for known actions (edit/new) but raises 'undefined method' for custom actions.


jachenry:
hold off on this. just pulled locally and tests are failing. I'd still like to get your opinion on my changes.  Maybe you can think of a better solution to the defect.

jachenry:
going to close this patch because it breaks too many tests.  I'll rethink a fix.

PR #202, opened 2012-01-30
failing test for custom actions with and without optional parents

jachenry:
I tried creating a patch but was unsuccessful. Could you please look at these failing tests to see if you have an idea for a workarount.  The problem roots from custom actions not having a polymorphic path. The workaround in [this post](http://stackoverflow.com/questions/6142013/polymorphic-path-for-custom-collection-route) worked for me on an individual basis.  Instead of using the custom action url helpers `archive_resource_url` I am using the main helper with a :action option `resource_url(:action => :archive)`

### DIFF
--- a/test/url_helpers_test.rb
+++ b/test/url_helpers_test.rb
@@ -119,6 +119,14 @@ class ButtonsController < InheritedResources::Base
   custom_actions :resource => :delete, :collection => :search
 end
 
+class PanelController < InheritedResources::Base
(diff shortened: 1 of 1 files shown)


rafaelfranca:
This issue has been automatically closed because of inactivity.

If you can still reproduce this error on the newest version, please reply with all of the information you have about it in order to keep the issue open.

Thank you.

PR #211, opened 2012-02-23
Document a case when you need to use the "inherit_resources" macro.

mforsyth:
Hi there - I came across a case where I needed to use the "inherit_resources" macro rather than inheriting from InheritedResources::Base, and I would have found it helpful if this type of situation was more verbosely described in the documentation.    So I've added an example of the type of scenario where "inherit_resources" comes in handy.


salzig:
+1

PR #226, opened 2012-04-23
Added :router to defaults

grandall:
If you have a controller in an application that inherits from a controller in a mountable engine (isolated namespace) and the child controller is also inheriting views from the parent controller, then using IR's url helpers (e.g. collection_path, resource_path) in the parent views will fail. The reason is that the url helpers need to be prepended with the name of the router (i.e. 'main_app').

To solve this problem, I added a :router parameter to the defaults function that allows the user to specify a string for the appropriate router (e.g. 'main_app') for a particular controller. The parameter almost exactly mirrors the :route_prefix parameter, so it's a pretty low risk fix.


joelmoss:
Any chance you can write some tests for this please?

joelmoss:
Please reopen this PR when you have tests. thx

PR #229, opened 2012-07-02
Memoize resource_params

### LINKED ISSUE #216: Memoizable resource_params
The resource_params method in InheritedResources::BaseHelpers is:

``` ruby
def resource_params
  rparams = [params[resource_request_name] || params[resource_instance_name] || {}] 
  rparams << as_role if role_given?
  rparams
end
```

A new hash is created when params is blank. However, this prevents generic manipulation of the parameters with before_filters, for example for setting the author of a post. This is because the hash is created again in build_resource

I think memoize resource_params would be valuable. What do you think?

atd:
Fixes #216

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -298,8 +298,13 @@ module InheritedResources
         url ||= root_url rescue nil
       end
 
-      # extract attributes from params
+      # memoize the extraction of attributes from params
       def resource_params
+        @resource_params ||= build_resource_params
+      end
+
+      # extract attributes from params
+      def build_resource_params
         rparams = [params[resource_request_name] || params[resource_instance_name] || {}]
         rparams << as_role if role_given?
         rparams


lib/inherited_resources/base_helpers.rb
```diff
@@ -298,8 +298,14 @@ def smart_collection_url
         url ||= root_url rescue nil
       end
 
-      # extract attributes from params
+      # memoize the extraction of attributes from params
       def resource_params
+        @resource_params ||=
+          build_resource_params
```
rafaelfranca: Put everything in the same line please.

atd:
Did you mean this way?

rafaelfranca:
Yes, Thanks. @joelmoss all yours now. ;)

joelmoss:
thx @atd

nestegg:
This change breaks my tests.  I can't find anywhere that resource_params ever gets called more than once, so I can't see the value in memoizing.  I rewrote my tests to get around this, but this change seems dangerous.

PR #230, opened 2012-07-02
Bump responders dependency version

jayzes:
Ran into some issues in a project that were fixed in a newer version of the responders gem. Not sure if there's a specific reason that this is locked back to the 0.6 series, but the tests pass and my app runs fine with the 0.9 series.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,7 +3,7 @@ PATH
   specs:
     inherited_resources (1.3.1)
       has_scope (~> 0.5.0)
-      responders (~> 0.6)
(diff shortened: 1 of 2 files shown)


carlosantoniodasilva:
@jayzes I think there's no special reason to lock this, but you should be able to update responders in your app with a `bundle update responders`, since inherited resources relies on `responders ~> 0.6` and not `~> 0.6.0`. In any case, I think it's fine to bump here anyway, if you say everything is passing ;)

joelmoss:
thx

PR #237, opened 2012-09-04
add support for controller-based param whitelisting (ala strong parameters)

rubysolo:
if the inherited resources controller has a method named <resource name>_params, it will be called to construct the params hash that will be used for create/update.

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -305,7 +305,7 @@ module InheritedResources
 
       # extract attributes from params
       def build_resource_params
-        rparams = [params[resource_request_name] || params[resource_instance_name] || {}]
+        rparams = [whitelisted_params || params[resource_request_name] || params[resource_instance_name] || {}]
         if without_protection_given?
           rparams << without_protection
         else
@@ -313,6 +313,11 @@ module InheritedResources
         end
 
         rparams
+      end
+
+      def whitelisted_params
+        whitelist_method = :"#{ resource_request_name }_params"
+        respond_to?(whitelist_method) && self.send(whitelist_method)
       end
 
       # checking if role given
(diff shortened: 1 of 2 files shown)


jodigiordano:
+1. @rubysolo the documentation of [strong parameters](https://github.com/rails/strong_parameters#strong-parameters) states that _Using a private method to encapsulate the permissible parameters is just a good pattern_, so what do you think of `respond_to?(whitelist_method, true)` instead of `respond_to?(whitelist_method)` ?

rubysolo:
:+1: I'm down with that.

garysweaver:
Sorry if this is way off-topic, but anyone have any experience with using InheritedResources with ActiveAdmin? @latortuga said that this pull request was adding support for StrongParameters in InheritedResources, but looking at it again, I don't see any calls to permit here and it appears to just be adding similar behavior to SP. Is support for SP being added to IR, and if it is, will it support the ability to permit _all_ parameters regardless of whether they relate to actual methods (attribute or otherwise), because that is something that AA needs, because of some password field related behavior it provides, as discussed here were a patch for SP in AA was being discussed (and AA uses IR heavily, so it would be better to do in IR): https://github.com/gregbell/active_admin/issues/1731 Thanks.

garysweaver:
Just saw: https://github.com/josevalim/inherited_resources/issues/236 - sorry for interrupting. Looks like calls to SP's permits are not getting added but that is something each person can put in themselves.

rubysolo:
You're right, this is not Strong Parameters proper, but the ability to do whitelisting in the controller _similarly_ to SP.  This patch should probably be revisited now that SP is more fully-baked and will be included in Rails 4.

garysweaver:
Yeah, since they are axing mass assignment security and SP isn't optional, we're going ahead and using Strong Parameters with Rails 3.2.8 (and ActiveAdmin which uses InheritedResources) and hitting pain head-on: https://github.com/rails/strong_parameters

joelmoss:
Closing this in favour of using strong parameters. Happy to accept a PR to integrate that. thx

PR #241, opened 2012-09-19
Incorrect `collection_path` is generated for uncountable shallow resources

dim:
The title says it all, nothing to add :)


dim:
Any news? This is really quite an annoying bug, even if not many other people seem to be affected by it.

joelmoss:
thx

PR #245, opened 2012-10-22
Ensure subclassed IR controllers can correctly resolve their resource_class

tardate:
Until recently, it has been possible to inherit from an IR-controller. Although this wasn't exactly documented or tested behaviour - it worked, and I've certainly used it quite often. But it broke in recent history.

These couple of commits restore the behaviour, and make sure there are tests to match.

NB: this reverts some changes in commit b21aa27 which allow for resource_class to be overwritten by a method in the controller class (which breaks the class inheritance behaviour and is also not asserted by any tests). 

It seems b21aa27 is possibly redundant(?), since resource_class may be over-ridden in defaults anyway. And it looks a little tricky to be able to allow overriding resource_class with a method _and_ allowing controller inheritance. I've tried to find out more about b21aa27 but haven't seen much discussion on the point yet - see https://github.com/josevalim/inherited_resources/commit/b21aa27b218aa831592a449e554a2361bfc39d6f#app/controllers/inherited_resources/base.rb

Hopefully this pull request will put the matter up for consideration one way or the other. It kind of boils down to choosing if IR should:
(a) allow overriding resource_class with a method OR
(b) continuing to support the ability to subclass IR controllers OR
(c) dig deeper to see if we can do both (a) and (b)

Since both (a) and (b) are actually undocumented behaviours at this point, I'm not actually sure what the "right" answer is. Anyone?

### DIFF
--- a/app/controllers/inherited_resources/base.rb
+++ b/app/controllers/inherited_resources/base.rb
@@ -30,8 +30,7 @@ module InheritedResources
                       :parent_url, :parent_path,
                       :smart_resource_url, :smart_collection_url
 
-        self.class_attribute :resource_class, :instance_writer => false unless self.respond_to? :resource_class
-        self.class_attribute :parents_symbols,  :resources_configuration, :instance_writer => false
+        self.class_attribute :resource_class, :parents_symbols,  :resources_configuration, :instance_writer => false
 
         protected :resource_class, :parents_symbols, :resources_configuration,
           :resource_class?, :parents_symbols?, :resources_configuration?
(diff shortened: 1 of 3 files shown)


bashmish:
+1

I have STI with three levels of inheritance and controllers for the last two of them.
`resource_class` is broken in the second case as it has been already set on the previous level.

Simply change

```
self.resource_class ||= begin
```

in first priority into

```
self.resource_class = begin
```

and everything works as expected.

Is this pull request be merged sometime? Or some exploration needed?

rafaelfranca:
This need a rebase.

PR #251, opened 2012-12-14
support create with three block arity

mcspring:
This pull request add ability for user wanted to access the new created resource.

With this pull request, you can write custom create as follow:

```
def create
  create! do |resource, success, failure|
    success.html { redirect_to url_for(resource) }
    failure.html { redirect_to root_url }
  end
end
```

I met this requirement when I use _acts-as-taggable-on_ gem, adding tag support for some entities. In that scene, I need to do thing likes `current_user.tag(resource, :on => tags, :with => params[:tags])`.

Hope this will useful.

Thanks.

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -233,6 +233,18 @@ module InheritedResources
 
       # Used to allow to specify success and failure within just one block:
       #
+      #   # three block arity
+      #   def create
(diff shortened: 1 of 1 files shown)


carlosantoniodasilva:
`resource` should be already available for you in that method, since it's a helper method in the controller. There shouldn't be necessary to add a third argument to the block. Thanks!

PR #255, opened 2013-02-08
added params[:ids] = [1, 2, 3] support to collection actions

heartsentwined:
Adds support for params[:ids] on collection actions, making the following URLs work as expected, while not breaking any existing functionality:
- `/users?ids[]=1&ids[]=2`
- `/users?ids=1`
- `/users`

This is a primitive support for batch API actions, currently I have only written a test for the `index` action, but since the change is in the underlying `collection` action, it should work for anything else that depends on `collection` too.

This pull request was inspired by batch API conventions that are emerging; specifically this was based on [ember.js](http://emberjs.com/)'s REST adapter, which maps a request of loading users with ID 1 and 2 to `GET /users?ids[]=1&ids[]=2`.

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -22,7 +22,9 @@ module InheritedResources
       def collection
         get_collection_ivar || begin
           c = end_of_association_chain
-          set_collection_ivar(c.respond_to?(:scoped) ? c.scoped : c.all)
+          coll = c.respond_to?(:scoped) ? c.scoped : c
+          coll = params[:ids] ? coll.find(params[:ids]) : coll.all
+          set_collection_ivar(coll)
         end
       end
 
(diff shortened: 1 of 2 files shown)


carlosantoniodasilva:
I'd rather not support it directly in IR, since it's nor IR responsibility and goal. Thanks for your contribution.

heartsentwined:
Does IR plan to support batch actions?

Use case will primary be in developing APIs, to reduce round-trips by performing batch requests in CRUD actions.

If so, I can help refactor the codebase.

joelmoss:
Yes, I would agree with @heartsentwined. Think this is not within the scope of IR and can be easily done using `end_of_association_chain`.

Thanks anyway.

carlosantoniodasilva:
This would be a good fit for a separated gem, though ;)

PR #194, opened 2012-01-08
Allow for nesting of singleton belong_to's.

DouweM:
In reference to issue #141.

Note that

``` ruby
class ThingAController < InheritedResources::Base
    belongs_to :thing_b, :singleton => true
end
```

used to mean there was only one `ThingA` (so `resource_path` == `/thing_bs/123/thing_a`), while it now means there is only one `ThingB` (so `resource_path` == `/thing_b/thing_as/123`)

To specify there is only one `ThingA`, you'll need

``` ruby
defaults :singleton => true
```

PR #258, opened 2013-02-15
Pull request #194

stephenprater:
This is just pull request #194 cleaned up enough to merge cleanly.


stephenprater:
discovered that the patch does not address the (admittedly a little bizzare) case of multiply nested singleton resources.  fixed it.

Issue #236, opened 2012-08-28
Integration with strong_parameters


dpmccabe:
Has anyone figured out a way to integrate Inherited Resources with the [Strong Parameters](https://github.com/rails/strong_parameters) gem, which might be making its way into Rails 4? Sure, you could overwrite the create and update actions on all of your controllers, but that would defeat the purpose of using Inherited Resources at all.

khelben:
You need to overwrite the `#resource_params` method in your controller.

One gotcha though, you'll need to wrap the result in an array like this:
Second gotcha, IR also calls this method on get #new where the params most off the time is empty and would raise an `ParamsRequiredError`.

``` ruby
def resource_params
    return [] if request.get?
    [ params.require(:employee).permit(:name, :age) ]
end
```

edit: update example, thx to @btucker
edit2: thx @anlek  for reminding me of the get #new issue

btucker:
@khelben I believe the method name in your example above should be `resource_params`.

khelben:
@btucker thx for spotting, updated the example

anlek:
This seems to raise a `Required parameter missing: page` on the new action when I have this set:

``` ruby
  def resource_params
    [ params.require(:page).permit(:title, :permalink, :content) ]
  end
```

khelben:
That means that your params hash is missing the required ':page' key.

anlek:
Understood, but the 'new' action doesn't require the params hash at all. (unless it fail validation)

khelben:
You're right. You've hit something I've encountered myself.

The thing is: inherited resources also calls #resource_params on a new action, where most of the times the params is empty.

I've updated the example with some more explanation

xxswingxx:
Hi, I've been trying to use inherited resources with the `strong_parameters` gem as seen here.
It almost works, the only problem is that when the parameters reach the `sanitize_for_mass_assignment`method from `ActiveModel::ForbiddenAttributesProtection` it is converted from `ActionController::Parameters` to `Hash`, so the method ignores it because it doesn't respond to the method `permit`. This causes that it allows mass assignment for unpermitted attributes if they are passed as a `Hash` instead an `ActionController::Parameters`. Is anyone having the same issue?

EDIT: Nevermind, it was a problem with mongoid and how it passes multiparameter attributes to the method. I've already fixed it.

aaronchi:
Hey guys,

[...]

PR #260, opened 2013-02-27
Integrate with strong_parameters

taavo:
See also #236

We now check for a method `permitted_params` in `build_resource_params`, and use it instead of `params` if present. We're expecting `permitted_params` to be implemented using strong_parameters' `permit` method, like this:

```
def permitted_params
  params.permit(:widget => [:permitted_field, :other_permitted_field]) 
end
```

Note that this doesn't work if you use strong_parameters' `require` method, because whereas the above example returns a cleaned copy of `params`, `params.require(:widget).permit(:permitted_field, :other_permitted_field)` returns a cleaned copy of `params[:widget]`.

### DIFF
--- a/README.rdoc
+++ b/README.rdoc
@@ -539,6 +539,16 @@ And then you can rewrite the last example as:
       end
     end
 
+== Strong Parameters
+
+If your controller defines a method named permitted_params, Inherited Resources will call it where it would normally call params. This allows for easy integration with the strong_parameters gem:
+
+    def permitted_params
+      params.permit(:widget => [:permitted_field, :other_permitted_field])
+    end
+
+Note that this doesn't work if you use strong_parameters' require method instead of permit, because whereas permit returns the entire sanitized parameter hash, require returns only the sanitized params below the parameter you required.
+
 == Bugs and Feedback
 
 If you discover any bugs, please describe it in the issues tracker, including Rails and Inherited Resources versions.
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -305,7 +305,8 @@ module InheritedResources
 
       # extract attributes from params
       def build_resource_params
-        rparams = [params[resource_request_name] || params[resource_instance_name] || {}]
+        parameters = respond_to?(:permitted_params) ? permitted_params : params
+        rparams = [parameters[resource_request_name] || parameters[resource_instance_name] || {}]
         if without_protection_given?
           rparams << without_protection
         else
--- a/test/strong_parameters_test.rb
+++ b/test/strong_parameters_test.rb
@@ -1,0 +1,34 @@
+require File.expand_path('test_helper', File.dirname(__FILE__))
+
+class Widget
+  extend ActiveModel::Naming
+end
+
+class WidgetsController < InheritedResources::Base
+end
+
+class StrongParametersTest < ActionController::TestCase
+  def setup
+    @controller = WidgetsController.new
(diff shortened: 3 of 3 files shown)


josemarluedke:
:+1:

brendon:
So weird how something like this is merged just when I needed it! :)

It might pay to add in the README that the `permitted_params` method should be protected and not private as otherwise it won't be found. The strong_parameters gem recommends a pattern whereby the allowed parameters are built in a private method.

Alternatively the check for the method could include the private space?

thedanbob:
Actually, it doesn't seem to work even when `permitted_params` is only protected. It isn't being called by IR in my app unless I move it to public.

brendon:
You're correct @danbob. I just checked my code and my `permitted_params` is in the public space too. I must have figured that out, changed it, and forgot about my comment here.

Wait, actually, I've noticed that I have it in the public space in my Active Admin controller block, but in my main app I was able to put it in the protected space and it works fine. You don't happen to be using Active Admin?

thedanbob:
Nope, just IR and CanCan. I second the suggestion to allow it to live in private or at least protected, so I can keep all my non-action controller methods separate.

brendon:
That's fairly strange. But yes, it seems more logical to allow for the recommended SP technique.

VorontsovIE:
Are there any reasons to use strong parameters without `#require`? May be this behavior should be fixed while not so much people use inherited_resources with rails 4?
I think it can be done simultaneously with sticking to standard rails 4 `#{resource_name}_params` method naming because it reduces number of error and simplified integration with such libraries as cancan and others. We can generate to methods: `#permitted_params` which works in exactly the same way it works now and `#{resource_name}_params` which overrides `#permitted_params` and allows `params#require` to be used.

taavo:
Without `#require`, strong parameters still prohibits fields other than those marked by `#permit`, which means someone can update `User#email` without being able to update `User#is_admin`, which is most of the point of strong_parameters.

I still believe that `#require` should throw an error if the indicated parameter isn't present, but not drill down, but that's an issue for the strong_parameters team. It looks like this [came up already](https://github.com/rails/strong_parameters/pull/117), and someone may have come up with a plausible solution, but it's since been reverted.

VorontsovIE:
@taavo I only say that it's simplier to allow one use both methods and use strong parameters exactly the same way they do it usually. When you refactor code to use inherited_resources - it's too simple to forget changing allowed parameters to use permitted_resource instead of, say user_params and remove `#permit` call in preference of deeper nesting which was working before. I just can't understand which profits this way of handling gives us comparing to default rails 4 way of handling params. The only benefit I can see - is that this method named the same for all controllers, but we can do both methods: `#user_params` (defined by user) and `#permitted_params` (defined and used in the gem) which will just delegate to `#user_params` if this method exists or fallback to `params[#{resource_name}]` (or probably raise an exception) if not.
My first concern here is compatibility of code. Because user won't need to change anything in rails4 scaffold code and in old code using inherited_resources too (`permitted_params` is just already redefined by the user and thus not trying to fallback to `user_params`). Also cancan and other libraries can rely on rails4 naming when trying to work with strong parameters.
I'll send a pull request soon to explain my idea with code more explicitly.

VorontsovIE:
@taavo Sent a pull request: https://github.com/josevalim/inherited_resources/pull/286

vitaly:
@joelmoss @taavo  How is this better then just overriding `resource_params` in your controller? it doesn't do anything useful when we use `strong_parameters` as we don't need roles in this case.

PR #262, opened 2013-02-27
Compatibility with Ruby 2.0

corny:
In Ruby 2.0 `respond_to?` has been changed to return false for protected methods. This results in `apply_scopes` not beeing called.

It can be fixed passing true as a second argument. Ruby 1.9 requires the second parameter to look for private methods. So this change does not break anything.

http://tenderlovemaking.com/2012/09/07/protected-methods-and-ruby-2-0.html


corny:
@josevalim Are there any plans to release an new version, or are you looking for a new maintainer?

joelmoss:
Duplicate of #263

PR #263, opened 2013-02-27
Compatibility with Ruby 2.0 (including smart redirects fix)

mehanoid:
Original pull request: https://github.com/josevalim/inherited_resources/pull/262

Fixed same problem with "respond_to?" for smart redirects correct work.


corny:
Oh thanks, I failed to notice the second call.

wyaeld:
+1, ran into this yesterday.  Is there a reason for not publishing a new version of the gem?

PR #273, opened 2013-04-13
Don't call :scoped if it's deprecated.

dougo:
This avoids a deprecation warning in Rails 4.


travisp:
I might be doing something wrong, but this causes "uninitialized constant InheritedResources::BaseHelpers::ActiveRecord" errors when running the tests for me.

dougo:
@travisp: Oops, it should check for the existence of ActiveRecord too.  Sorry about that.  I'll try to submit another pull request this evening, unless someone else beats me to it.

PR #279, opened 2013-04-25
Unsplatting simplified

VorontsovIE:


joelmoss:
Please explain what this does and why it is needed? thx

VorontsovIE:
It just pass unsplatted parameters separate from options. `x=['a','b']; f(*x,'c')` does the same as `x=['a','b']; f(*(x<<'c'))` if you don't care about value of x. Because arguments for unsplatting (`options` and result of `with_chain`) are temporary arrays in your case, this modifications makes nothing except increasing readability.

PR #286, opened 2013-05-20
strong parameter support refined to allow standard Rails 4 notation

VorontsovIE:
This commit allows one to use standard rails 4 `params.require().permit()` notation.
It's done in such a way that this doesn't clash with `permitted_params` if it's already defined. But if it doesn't, then default `permitted_params` falls back to a permitting method with resource-specific name.
Method user should define is the same as scaffold generates and thus the same as other gems such as cancan will try to handle.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,3 +5,7 @@ gemspec
 gem 'rails', '>= 3.2', '< 5'
 gem 'mocha'
 gem 'turn'
+
+group :test do
+  gem 'strong_parameters'
+end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -80,6 +80,10 @@ GEM
       actionpack (>= 3.0)
       activesupport (>= 3.0)
       sprockets (~> 2.8)
+    strong_parameters (0.1.4)
+      actionpack (>= 3.2.0)
+      activemodel (>= 3.2.0)
+      railties (>= 3.2.0)
     thor (0.18.1)
     thread_safe (0.1.2)
       atomic
@@ -98,4 +102,5 @@ DEPENDENCIES
   inherited_resources!
   mocha
   rails (>= 3.2, < 5)
+  strong_parameters
   turn
--- a/README.md
+++ b/README.md
@@ -706,6 +706,16 @@ def build_resource_params
 end
 ```
 
+
+Instead you can stick to a standard Rails 4 notation (as rails scaffold generates) and write:
+
+    def widget_params
+      params.require(:widget).permit(:permitted_field, :other_permitted_field)
+    end
+
+In such case you should remove #permitted_params method because it has greater priority.
+
+
 ## Bugs and Feedback
 
 If you discover any bugs, please describe it in the issues tracker, including Rails and InheritedResources versions.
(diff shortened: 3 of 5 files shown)


travisp:
I think it's a good idea, but the tests should probably test the "required" support. Some quick testing seems to show that it works fine for the "update" controller action.. but it fails for me on the "new" controller action: a 'Required parameter missing' error occurs on the new action, which isn't normal rails behavior. The new action builds a resource and calls resource_params.

VorontsovIE:
@travisp Thank you for a response! Actually I didn't tested it with strong_parameters, and did only testing with stubs. I'll try to fix it.

VorontsovIE:
@travisp Can you please show code which caused the failure? In my tests everything works.

travisp:
@prijutme4ty Your code testing the new action should be "get :new, {}" because the new action typically does not have any parameters passed to it.

This test fails:

``` ruby
  def test_resource_params_from_new
    get :new, {}
    assert_response :success
    assert assigns(:widget)
  end
```

Results in: "Expected response to be a <:success>, but was <400>" and response.body after the get request contains "Required parameter missing: hotel"

VorontsovIE:
@travisp Understood the problem. I believe, it's due to inherited_resources using strong_parameter sanitization on new action, while it's not a common practice. I'll try to distinguish new and create action when call `#build_resource`

rafaelgoulart:
I'm having the same problem with the new action.

travisp:
@rafaelgoulart The code still hasn't been updated to deal with it. I might take a shot in a few weeks if nobody else gets to it: it would be very nice if inherited_resources could work with standard strong_parameters syntax and would also allow the use of "requires".

taavo:
I've come to the conclusion that the API I used in #260 was stupid, and this one is much better. I'll help out if my schedule opens up.

dim:
Any news? Would love to see this merged.

rolftimmermans:
This proposed API makes so much more sense. The current strong parameters solution is pretty awkward to use correctly.

StrangeElement:
It would be pretty helpful to add a note to the README meanwhile.

As I understand it (after some testing and debugging inside IR), permitted_params must not use params.require(:model).permit(...), but rather params.permit(...). When require is set, IR receives empty parameters from BaseHelpers.build_resource_params so the model is empty.

aaronchi:
+1

To address the new issue, you can skip building params if request.get?

evserykh:
:+1:

dreoliv:
:up:

Any chance of merge for this PR?

joelmoss:
If @prijutme4ty can make the PR merge cleanly, I'd be happy to merge it.

killthekitten:
Ping @prijutme4ty

VorontsovIE:
@travisp @joelmoss Tried to resolve issue with new action by rescuing an exception. More detailed explanation in a comment before `#permitted_params` method. Is it acceptable?
Everybody, sorry for such a long delay.

joelmoss:
LGTM!

dreoliv:
:heart:

PR #287, opened 2013-05-21
Fixes tests which failed due to absence of ActiveRecord in tests.

VorontsovIE:
Made it possible to test in existence or absence of `ActiveRecord::DeprecatedFinders`

Before this patch some of my tests failed due to check (`ActiveRecord.const_defined?(:DeprecatedFinders)`) of deprecated finders existence in `BaseHelpers#collection`.
Now one can explicitly set whether he wants to test in existence or absence of deprecated finders.
Now all tests doesn't use deprecated finders (and if turn it on they fail because appropriate methods were not defined). Probably it's wise to write also tests in case that deprecated finders defined.

### DIFF
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -44,3 +44,18 @@ class ActiveSupport::TestCase
     @routes = InheritedResources::Routes
   end
 end
+
+# make possible testing with existence or absence of deprecated finders
+class ActiveRecord
+  class << self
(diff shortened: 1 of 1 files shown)


dougo:
I think this issue was fixed by this pull request: https://github.com/josevalim/inherited_resources/pull/288

VorontsovIE:
@dougo I think it was. I'll close the issue

PR #288, opened 2013-05-24
Use a safer check for :scoped being deprecated.

dougo:
This works even if ActiveRecord is not defined.

Fixes the test failure mentioned here: https://github.com/josevalim/inherited_resources/pull/273#issuecomment-18204924

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -22,7 +22,7 @@ module InheritedResources
       def collection
         get_collection_ivar || begin
           c = end_of_association_chain
-          use_scoped = c.respond_to?(:scoped) && !ActiveRecord.const_defined?(:DeprecatedFinders)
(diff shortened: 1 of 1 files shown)


dim:
That didn't really do the trick. Calling `Model.all` returns an `ActiveRecord::Relation`, while calling `parent.children.all` returns an `Array`, which causes inconsistencies down the line. I also filed a ticket with Rails (https://github.com/rails/rails/issues/11310), but I think we may need a fix in inherited resources too, how about:

``` ruby
def collection
  get_collection_ivar || begin
    c = end_of_association_chain
    s = if c.respond_to?(:scoped) && !defined?(ActiveRecord::DeprecatedFinders)
      c.scoped
    elsif c.is_a?(ActiveRecord::Relation)
      c
    else
      c.all
    end
    set_collection_ivar(s)
  end
end
```

PR #301, opened 2013-08-08
predicate methods for action type #collection_action? and #resource_action?

VorontsovIE:
new controller methods #resource_action? and #collection_action? which can be used for example in conditional filters to decorate resource but not collection.

```
before_filter :if => :resource_action? do
  @book = @book.decorate
end
```


voondo:
+1

VorontsovIE:
Any comments on this feature?

rafaelfranca:

PR #315, opened 2013-09-09
fix for resource_class default value definition

evilguc:
This fix solves problems with default value definition for resource_class accessor.

Example:
we have model

``` ruby
Controller::User
```

and Controller for it:

``` ruby
Controller::UsersController
```

Old behavior causes wrong definition of resource_class:

``` ruby
'Controller::UsersController'.sub(/Controller/, '').singularize # 'UsersController'
'UsersController'.constantize # raises exception and resource_class becomes nil
```

My solution is to substitute 'Controller' postfix from controller name. So Controller::UsersController matches Controller::User model


evilguc:
I moved it to https://github.com/josevalim/inherited_resources/pull/348.
Closed

PR #317, opened 2013-09-14
Example how to include pagination data in the response when creating an ...

koloritcm:
...API


rafaelfranca:
I don't think this belongs to this project. Thank you for the pull request.

PR #318, opened 2013-09-18
Don't generate helper methods when resource's name is Resource

ka8725:
I have troubles when I inherit `ResourcesController` from `InheritedResources::Base`: I try to use `resource_path` or `edit_resource_path` but application gets infinitive recursion and as a result I have exception:

``` ruby
SystemStackError - stack level too deep:
  (gem) activesupport-3.2.12/lib/active_support/core_ext/array/extract_options.rb:23:in `'
```

So I see here only one solution - don't define helper methods when resource's class name is "Resource". This fix is about it

### DIFF
--- a/lib/inherited_resources/url_helpers.rb
+++ b/lib/inherited_resources/url_helpers.rb
@@ -216,6 +216,9 @@ module InheritedResources
       prefix = prefix ? "#{prefix}_" : ''
       ivars << (ivars.empty? ? 'given_options' : ', given_options')
 
+      # Don't define helpers if resource's name is "Resource" to prevent collisions
+      return if resource_class.to_s == 'Resource' && name != :collection
+
       class_eval <<-URL_HELPERS, __FILE__, __LINE__
         protected
           def #{prefix}#{name}_path(*given_args)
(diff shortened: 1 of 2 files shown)


avsej:
Do not use `Resource` to name your resources

ka8725:
Rails doesn't prohibit to name model with `Resource`. The gem is extension for rails, so it should not break rails conventions

avsej:
all correct. see definition of SHOULD
http://www.ietf.org/rfc/rfc2119.txt

>   SHOULD   This word, or the adjective "RECOMMENDED", mean that there
>   may exist valid reasons in particular circumstances to ignore a
>   particular item, but the full implications must be understood and
>   carefully weighed before choosing a different course.

ka8725:
Ok, I see that this problem can be fixed with removing method. It means that if you have collision like this you always can delete generated helper methods from the controller:

``` ruby
class ResourcesController < InheritedResources::Base
  # HACK: delete generated method by inherited resources.
  # It needs to use real generated method for actual resource
  remove_method :resource_path
end
```

So it's not necessary to change `inherited_resources` code

PR #326, opened 2013-10-09
Unnecessary modification of the class_name in belongs_to breaks class name

cheef:
I think that there is no need to modify string in class name `:class_name => 'Foo'` in `belongs_to`. According to the comments it was added for compability with **ActiveRecord** and here **belongs_to** only constantized.

Also it's destructive modification, it brakes class name in cases like:

``` ruby
'Library::Base'.to_s.pluralize.classify # => 'Library::Basis
```

Test added.

### DIFF
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -175,7 +175,7 @@ module InheritedResources
           config[:parent_class] = options.delete(:parent_class) ||
             begin
               class_name = if options[:class_name]
-                options.delete(:class_name).to_s.pluralize.classify
+                options.delete(:class_name)
               else
                 namespace = self.name.deconstantize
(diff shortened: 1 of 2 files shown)


rafaelfranca:
Could you rebase this pull request and see if this is still needed?

sevaorlov:
@rafaelfranca, yes, the issue is still present in master

varyonic:
@cheef IR has new maintainers.  Please rebase on the current master if interested.  If not then someone may rebase for you.

rafaelfranca:
@varyonic maybe it would be better if you merge it manually?

varyonic:
Merged manually, see https://github.com/activeadmin/inherited_resources/commit/65a4cea1fb2e58a70abd3f6b00f632b3845543c9

rafaelfranca:
Thanks!

PR #330, opened 2013-10-17
fix belongs_to for models without namespace in namespaced controllers

Bonias:
Bug was introduced by this commit https://github.com/josevalim/inherited_resources/commit/630a2775fab351540fc9bd8aa0ce43d4d039f44a, and brake inherited_resources in such situation:

```
class Group < ActiveRecord::Base
end 
class User < ActiveRecord::Base
end 
module Admin
  class Users < InheritedResources::Base
    belongs_to :group
  end
end
```

### DIFF
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -186,13 +186,12 @@ module InheritedResources
                   if klass.safe_constantize
                     break
                   else
+                    klass = nil
                     namespace = namespace.deconstantize
                   end
                 end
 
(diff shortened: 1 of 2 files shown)


cheef:
Does my request fix this issue? #326

Bonias:
It does not. Your modification concerns belongs_to with class_name option.

My pull request fixes belongs_to without class_name or parent_class option. In my example resources_configuration[:group][:parent_class] should equal Group, but is nil.

rafaelfranca:

PR #337, opened 2013-11-17
Added ability to change the default redirect order for actions

efexen:
Hey, I've been using inherited resources a great deal lately and find myself constantly changing the redirect path after successful create or update. Came up with this nifty idea for allowing the order of the redirects for CUD to be overridden which saves you from having to defined each action with the short hand redirect. 

Could you have a look through the diff below and merge if you think it suitable, I've tried to keep all interfaces same and followed the styles and conventions when possible. I've added some new tests as well to ensure this new functionality is correct and amended the README with instructions on how to use this.

Thanks
.FxN

### DIFF
--- a/README.md
+++ b/README.md
@@ -368,6 +368,21 @@ end
 
 This controller redirect to parent window after all CUD actions.
 
+These default orders can be overridden using redirect_order and passing either a single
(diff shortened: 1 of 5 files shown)


rafaelfranca:

PR #339, opened 2013-11-21
checkout "association_chain" method for default behavior

ilya-manin:
Hello, dudes! :rage2:
I have a little trouble with nested controllers and singleton option.

There are my models:

``` ruby
class Speciality < ActiveRecord::Base
  has_many :subspecialities
end

class Subspeciality < ActiveRecord::Base
  belongs_to :speciality
  has_one :work_plan
end

class WorkPlan < ActiveRecord::Base
  belongs_to :subspeciality
end
```

and their controllers:

``` ruby
class SpecialitiesController < ApplicationController
  inherit_resources
end

class SubspecialitiesController < ApplicationController
  inherit_resources
end

class WorkPlansController < ApplicationController
  inherit_resources

  defaults :singleton => true

  belongs_to :speciality do
    belongs_to :subspeciality
  end
end
```

When I visit specialities/1/subspecialities/2/work_plan/new I got an error:

``` ruby
 undefined method `specialities' for #<Subspeciality:0x00000000000000>
```

I found on 138 line of https://github.com/josevalim/inherited_resources/blob/master/lib/inherited_resources/base_helpers.rb#L135-L147 symbols_for_association_chain reversed if singleton option is true.
Why was symbols_for_association_chain reversed? I can rewrite belongs_to in my controller in reverse order but why single option change default behavior?

PS

I checkouted the association_chain method to state before this commit 7f455095f55f1b9372ac28ce4e5cd8679b93a231 (https://github.com/josevalim/inherited_resources/commit/7f455095f55f1b9372ac28ce4e5cd8679b93a231#diff-9e560d176d0aba6271a143b4b0631e1d).
All tests are passed.

### DIFF
--- a/test/nested_singleton_test.rb
+++ b/test/nested_singleton_test.rb
@@ -29,17 +29,17 @@ end
 # /party/37/venue/address case
 class AddressController < InheritedResources::Base
   defaults :singleton => true
-  belongs_to :venue, :singleton => true do
-    belongs_to :party
+  belongs_to :party do
+    belongs_to :venue, :singleton => true
   end
 end
 
 #and the more pathological case
 class GeolocationController < InheritedResources::Base
   defaults :singleton => true
-  belongs_to :address, :singleton => true do
+  belongs_to :party do
     belongs_to :venue, :singleton => true do
-      belongs_to :party
+      belongs_to :address, :singleton => true
     end
   end
 end
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -133,17 +133,10 @@ module InheritedResources
       # current resource).
       #
       def association_chain
-        @association_chain ||= begin
-          symbol_chain = if resources_configuration[:self][:singleton]
-            symbols_for_association_chain.reverse
-          else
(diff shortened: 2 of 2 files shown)


evserykh:
:+1: default behavior is more expected

test/nested_singleton_test.rb
```diff
@@ -29,17 +29,17 @@ class VenueController < InheritedResources::Base
 # /party/37/venue/address case
 class AddressController < InheritedResources::Base
   defaults :singleton => true
-  belongs_to :venue, :singleton => true do
-    belongs_to :party
+  belongs_to :party do
+    belongs_to :venue, :singleton => true
```
rafaelfranca, line 33: Why this was changed?

aratak, line 33: It should be not reversed as @XaMuT described. Got the same issue and was surprised too.

Fivell (approved):
@varyonic according to docs here https://github.com/activeadmin/inherited_resources#nested-belongs-to
 @xamut  is right , it should be declared starting from root resource , so thats why tests were changed, after all I agree with this change

PR #341, opened 2013-12-11
When parent exists, make sure the created resource is associated

fredwu:
Perhaps I'm missing something really obvious, but I was surprised to find that when a resource has a parent, it does not get created under the parent.

This PR makes sure that a child object is correctly associated with its parent upon its creation.


rafaelfranca:
This changes the behaviour and may break a lot of applications. That said I prefer to not merge since we are in maintenance only mode.

PR #344, opened 2014-01-04
Revert "Unsplatting simplified"

sodabrew:
This reverts PR #279 commit bf659843dccec5b72b105e6ff9a581a84b9d53ef, which broke Ruby 1.8.7 compatibility.


joelmoss:
We don't support Ruby anymore 1.8.

sodabrew:
You dropped support on a micro release. That is irresponsible at a minimum.

PR #351, opened 2014-03-04
Parent scopes added

Sharangovich:
Useful option to work with different scopes.
Fixes situation, when in a parent controller resource has limited scope, but you can access nested resource through direct url like _localhost:3000/projects/289/comments/new_.


joelmoss:
Tests please

PR #356, opened 2014-04-01
Changed default parent class of InheritedResources::Base.

DiMarcello:
I always had this issue with namespaced applications, but that could be easily overcome by defining an pointer constant to the namespaced ApplicationController.

Now I have a namespaced Rails Application, that depends on an namespaced Rails Engine, that uses active_admin, which uses inherited_resources. And none, including the application itself have an ApplicationController in the root.

Also I would raise the discussion about moving Base to the lib directory, and have maybe an InheritedResources::ApplicationController < ::ApplicationController; inherit_resources; end 
if one needs it.

I don't really understand your architecture of tests, otherwise I would have written one for this change.
For backwards compatibility I would suggest an generator or something that creates an initializer. 

Greetings.

### DIFF
--- a/app/controllers/inherited_resources/base.rb
+++ b/app/controllers/inherited_resources/base.rb
@@ -8,7 +8,11 @@ module InheritedResources
   # call <tt>default</tt> class method, call <<tt>actions</tt> class method
   # or overwrite some helpers in the base_helpers.rb file.
   #
-  class Base < ::ApplicationController
+  # By default the base class inherits from ActionController::Base, since it is not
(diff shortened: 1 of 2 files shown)


joelmoss:
Nice, but the default parent controller should still be `::ApplicationController` in order to not break anything.

robertomiranda:
100% agree :+1:. is really annoying inherit everything from ApplicationController. Also agree with @joelmoss pitifully we need to keep `::ApplicationController` as default parent

rafaelfranca:

PR #362, opened 2014-05-07
Undefine url helpers if actions excluded

onemanstartup:
For example i want to create controller without 2 actions. And I have another controllers but with those actions. 

``` ruby
class Admin::UsersController < AdminController
  inherit_resources
  actions :all, except: [:create, :new]
  # ...
end
```

All controllers share common partial

``` ruby
    - if controller.respond_to?(:new_resource_url, true)
      span= link_to '+', new_resource_url
```

Without undefining url helpers I can't call respond_to?(:new_resource_url) and it seems not smart, having methods(url helpers) witch don't work.

Feel free to close or suggesting another way to do it.
Thanks.


rafaelfranca:

PR #366, opened 2014-06-02
Update responders dependency from 1.0.0 to 1.1.0

lksv:


inherited_resources.gemspec
```diff
@@ -19,6 +19,6 @@ Gem::Specification.new do |s|
   s.test_files    = `git ls-files -- {test,spec,features}/*`.split("\n")
   s.require_paths = ["lib"]
 
-  s.add_dependency("responders", "~> 1.0.0")
+  s.add_dependency("responders", "~> 1.1.0")
```
rafaelfranca, line 22: Could you use `~> 1.0` only? Since we are in the 1.x release the API is stable and all 1.x releases are compatible. Doing this way we don't need to change the gemfile every release of responders.

rafaelfranca, line 22: I'll do

Issue #325, opened 2013-10-09
belongs_to and undefined method `find' for nil:NilClass


MrMoins:
Hi,

I use ActiveAdmin with Rails4 and Ruby 2.0.0-P247 and I encounter a problem apparently related to inherited_resources.

My situation is described here [#2518](https://github.com/gregbell/active_admin/issues/2518)  and clean rails app is [available here](https://github.com/gregbell/active_admin/issues/aa-belongs-to) to reproduce the bug.

MrMoins:
(bump)

polymorfiq:
If this helps:

I experienced this problem when switching from the default gem commit:
https://github.com/josevalim/inherited_resources/commit/ae1516a3f1a01f98beac3334f76831a1c793c521

to the latest commit. Doing a checkout of that previous commit fixed my issue.

lksv:
Same problem. 

It happens on belongs to association e.g.:

```
 module Public
   class SubscriptionOrdersController < Public::ResourcesController
     belongs_to :magazine_group, optional: true

     def build_resource
       build_resource! do |r|  #<--- falls here
       #....
```

Digging deep I find out that this issue is cased by regression  in #303 (commit https://github.com/josevalim/inherited_resources/commit/630a2775fab351540fc9bd8aa0ce43d4d039f44a ). The exception is raised (uninitialized constant Public::MagazineGroup) but it is silently ignored by the rescue block in the referenced commit.

I did a quick fix of this: https://github.com/lksv/inherited_resources/commit/93640c35200fe59f3b8d93fb57da3e0e2916d3d5 but I am not putting it as pull request because of tests needs to be extended and I am not sure of if it solves situations.

amiel:
Just to cross-reference, I just recently helped someone troubleshoot this issue: http://stackoverflow.com/questions/24958503/belongs-to-on-activerecord-rendering-as-not-found/25028452

rafaelfranca:
Fixed by https://github.com/josevalim/inherited_resources/pull/376

PR #367, opened 2014-06-05
 #325 fix: set belongs_to properly for namespaced controller and non_namespaced model

lksv:
this pull request fix issue #325
This issue occurred when controller is in namespace but use model without namespace.

### DIFF
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -170,6 +170,7 @@ module InheritedResources
           self.resources_configuration[:self][:shallow] = true if shallow
 
           config = self.resources_configuration[symbol] = {}
+          config[:singleton]       = options.delete(:singleton) || false
 
           class_name = ''
           config[:parent_class] = options.delete(:parent_class) ||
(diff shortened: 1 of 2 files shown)


pbhogan:
:+1:

aaronchi:
:+1:

rohitpaulk:
+1

phlegx:
:+1:

aaronchi:
would be really nice if this could be merged and a new version released. It is holding up other gems like active_admin from using the latest version of IR

jasonayre:
:thumbsup:  +1

@josevalim :)

bronislav:
There are possible problem with `klass.safe_constantize`. In my environment it returns correct result starting from second call. First call always returns wrong result. I will come back with MVE if I will found why it's happens.

cknoxrun:
I've tested this with a whole bunch of controllers in my large app, all seems to work

petergoldstein:
This seems to be relevant to an issue that is locking ActiveAdmin to IR 1.4.x.  You can see the discussion here - https://github.com/activeadmin/activeadmin/pull/3193

rafaelfranca:

PR #369, opened 2014-06-14
Test case for PR #367

rohitpaulk:
Tests that `belongs_to` assigns the `parent_class` properly for a namespaced controller and a non-namespaced model.


rafaelfranca:

PR #378, opened 2014-10-23
change controller scaffold template in order to build resource params

dmitri-wm:
Change scaffold controllers template in order to generate resource params.

### DIFF
--- a/lib/generators/rails/templates/controller.rb
+++ b/lib/generators/rails/templates/controller.rb
@@ -4,10 +4,10 @@ class <%= controller_class_name %>Controller < InheritedResources::Base
 <% end -%>
 
 private
-
-  def <%= singular_name %>_params
-    params.require(:<%= singular_name %>).permit(<%= attributes_names.map{ |a_name| ":#{a_name}" }.join(", ") %>)
-  end
-
+  <% if Rails::VERSION::MAJOR >= 4 || defined?(ActiveModel::ForbiddenAttributesProtection) %>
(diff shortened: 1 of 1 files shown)


lib/generators/rails/templates/controller.rb
```diff
@@ -2,4 +2,12 @@ class <%= controller_class_name %>Controller < InheritedResources::Base
 <% if options[:singleton] -%>
   defaults :singleton => true
 <% end -%>
+
+private
+
+  def <%= singular_name %>_params
+    params.fetch(:<%= singular_name %>, {}).permit(<%= attributes_names.map{ |a_name| ":#{a_name}" }.join(", ") %>)
```
rafaelfranca: This fetch will fail in case the singular_name key is not present. I believe we should change it to `require`

rafaelfranca: I mean the fact of the hash with key with the same name of singular_name not being present in the params hash. If that is the case the fetch will return a Hash instance that doesn't responds to `premit` and will failt.

dmitri-wm: Yes, I've fixed it at the next commit
https://github.com/dmitri-wm/inherited_resources/commit/1c044c62cee66be8dd6dc4bc5dd18b1179af78aa

rafaelfranca:
I think we should only generate it if the application is a Rails 4 application or it is using strong parameters right?

dmitri-wm:
sure, done

PR #382, opened 2014-11-13
Fix url helpers on mountable engines

juniorz:
This is a rebase of #191.


juniorz:
@rafaelfranca any update on this?

Issue #381, opened 2014-11-10
Rails 4.2


timoschilling:
@rafaelfranca Thanks agin for your [announcement](https://github.com/activeadmin/activeadmin/pull/3193#issuecomment-62421649) to help with `inherited_resources` and Rails 4.2.

@andriytyurnikov says on https://github.com/josevalim/inherited_resources/issues/373#issuecomment-54052409 that he works on a fix ([andriytyurnikov/inherited_resources branch rails_4_2](https://github.com/andriytyurnikov/inherited_resources/tree/rails_4_2)), that I have tried  some days a go and the basics themes to work.

And @pedrofaria says (https://github.com/activeadmin/activeadmin/issues/3346#issuecomment-62421948) that he has a fork ([codecraft63/inherited_resources](https://github.com/codecraft63/inherited_resources)) to support Rails 4.2 too

timoschilling:
@rafaelfranca any updates on this, or can I help in some way?

rafaelfranca:
I already create a branch but I need to understand why the tests are failing randomly. https://github.com/josevalim/inherited_resources/tree/rails-4-2

boxofrad:
@rafaelfranca is this one of the failures you're getting?

```
ShowActionBaseTest#test_controller_should_render_show:
NoMethodError: undefined method `show' for #<UsersController:0x007fb9acb356d8>
    /Users/daniel/.rbenv/versions/2.1.1/lib/ruby/gems/2.1.0/gems/actionpack-4.1.5/lib/action_controller/metal/implicit_render.rb:4:in `send_action'
    /Users/daniel/.rbenv/versions/2.1.1/lib/ruby/gems/2.1.0/gems/actionpack-4.1.5/lib/abstract_controller/base.rb:189:in `process_action'
    /Users/daniel/.rbenv/versions/2.1.1/lib/ruby/gems/2.1.0/gems/actionpack-4.1.5/lib/action_controller/metal/rendering.rb:10:in `process_action'
    /Users/daniel/.rbenv/versions/2.1.1/lib/ruby/gems/2.1.0/gems/actionpack-4.1.5/lib/abstract_controller/callbacks.rb:20:in `block in process_action'
    /Users/daniel/.rbenv/versions/2.1.1/lib/ruby/gems/2.1.0/gems/activesupport-4.1.5/lib/active_support/callbacks.rb:82:in `run_callbacks'
[...]

PR #386, opened 2014-12-24
Fix Specs - Issue #381

scudelletti:
We remove the show method when we call .send(:actions, :all, :except => :show) as we can see [here](https://github.com/josevalim/inherited_resources/blob/master/lib/inherited_resources/class_methods.rb#L78-L80).
So we can't call `actions` twice to redo the method, which I think is really fair.

This PR fixes the specs which sometimes want the method back. 

https://github.com/josevalim/inherited_resources/issues/381


ashanbrown:
Rather than restoring the classes to their original state, I've proposed just using dynamic classes at https://github.com/josevalim/inherited_resources/pull/387.

scudelletti:
Awesome! o/

PR #387, opened 2014-12-29
allow test cases to safely modify controller class by making it unique p...

ashanbrown:
I believe the order related problems running the tests in rails 4.2 are due to test cases modifying the controller class.  This pull request creates a different controller class per test case, although I did have to define `name` to make this work (because it appears to be hard to dynmically create controller classes with a known name at the time that inherited resources as initialized).  An alternative might have been to just modify the singleton class in the testcases, but I think that's more confusing.

### DIFF
--- a/test/base_test.rb
+++ b/test/base_test.rb
@@ -12,7 +12,11 @@ class UsersController < AccountsController
   respond_to :js, :only => [:create, :update, :destroy]
   attr_reader :scopes_applied
 
-protected
+  def self.name
+    "UsersController"
+  end
+
+  protected
 
   def apply_scopes(object)
     @scopes_applied = true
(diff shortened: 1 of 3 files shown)


scudelletti:
I really liked this approach, its seems cleaner and it protects all test cases. :purple_heart:

When the class is inherited the name is invoked here, that's why you have to declare the method. :(
https://github.com/josevalim/inherited_resources/blob/master/lib/inherited_resources/class_methods.rb#L392-L396

rafaelfranca:
Can we dynamic declare the controller only in the test that need to change it?

ashanbrown:
A large fraction of the tests send messages that trigger changes in the controller.  I hadn't looked closely, but it appears most of them just appear to reset it afterwards.  Frankly, I'd rather just delete all the code that tries to reset the controller and always dynamically define the class.  Otherwise, I'd just go back to @scudelletti original proposal.

ashanbrown:
I've made changes to make it pass on rails 4.2.  Can't say actually say it works, but the tests pass against rails 4.2 gems now.

ashanbrown:
Cool.  I'm trying to get it to run on travis btw, with multiple rails versions and rubies.  Would you be open to a separate pull request for that?

rafaelfranca:
to get it working on travis @josevalim need to enable it first. I'll try to get him to enable it.

ashanbrown:
I'm just doing it on my own branch.  I'll let you know if I succeed.

josevalim:
Done!

PR #388, opened 2015-01-05
travis support for rails 4.1 and rails 4.2 builds

ashanbrown:
This is the .travis.yml file and additional Gemfile I used to get this to run on travis (and pass).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 .bundle
+test/gemfiles/*.lock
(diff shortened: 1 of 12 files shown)


rafaelfranca:
We need tests to work with 3.2/4.0 too.

ashanbrown:
OK, I'll add those too.  BTW, if you just pull this, I expect that the existence of a .travis.yml file will just cause travis to automatically pick it up for github account josevalim.

Curiously, the tests appear to fail on rubinious.  I don't use rubinious myself so I'm not sure I'll be looking more into it.  Maybe it's just some issue with the mocking library.

timoschilling:
:+1:

PR #389, opened 2015-01-05
run specs on travis against all major rails versions

ashanbrown:
Here is a travis.yml and associated files to run the tests for rails 3.2, rails 4.0, rails 4.1 and rails 4.2.  These should all pass.  I also enabled a bunch of ruby versions.  Note that rubinious appears to have a bug above version 2.2.10 that breaks inherited resources specs and behavior (for details see the bug I've created at https://github.com/rubinius/rubinius/issues/3272).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,13 @@ rvm:
   - 2.1.5
   - 2.2.0
   - jruby-19mode # JRuby in 1.9 mode
-  - rbx-2.1.1
+  - rbx-2.2.10
 gemfile:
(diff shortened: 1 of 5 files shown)


timoschilling:
:+1:

varyonic:
@ashanbrown @timoschilling I believe Rubinius 2.2 is now obsolete and this issue remains unresolved.  Given travis-ci/travis-ci/issues/5294 I would like to drop any efforts to support Rubinius.

PR #390, opened 2015-01-22
Removed HttpCacheResponder, closes #342

### LINKED ISSUE #342: Not using Responders::HttpCacheResponder
Hi there, I think:

`include Responders::HttpCacheResponder`

is to blame for the weird caching errors I'm seeing. Basically the contents of index and show views are cached (including flash messages) until I forcefully fetch the page again.

How would I go about overriding this?

bitboxer:
This responder hit us @bitcrowd and @brendon by surprise. It only adds 
caching if used as an API endpoint, not when serving html pages.

As discussed in the ticket #342 we are now removing it in the `InheritedResources::Responder`

### DIFF
--- a/lib/inherited_resources/responder.rb
+++ b/lib/inherited_resources/responder.rb
@@ -1,6 +1,5 @@
 module InheritedResources
   class Responder < ActionController::Responder
     include Responders::FlashResponder
-    include Responders::HttpCacheResponder
   end
 end


rafaelfranca:
Just to make sure I got the problem, this responders is making the API response to be cached, or it is the HTML response? For what I got in the original report at #342:

> Basically the contents of index and show views are cached (including flash messages) until I forcefully fetch the page again.

It seems it is blaming this responders to caching HTML pages.

bitboxer:
Sorry, was not detailed enough.

If you look into the [HttpCacheResponder](https://github.com/plataformatec/responders/blob/master/lib/responders/http_cache_responder.rb) you will see that this thing sends a `head :not_modified` if the browser sends a last modified that is the same as the resources `updated_at`. BUT only if the response is using `to_format`, which usually means API requests.

For every other request it is not cached and does not return a `head :not_modified`. This is a surprising behavior.

rafaelfranca:
Well, it is documented, so it is not surprising :smile:.

But I agree we should not include by default at inherited_resources.

bitboxer:
I could not find it when searching for "cache" on the readme of inherited_resources. Because of that it surprised me :wink:

nikolai-b:
Yep, this also surprised me. Any progress on this PR or if you are keeping it it would be nice to add it to the inherited_resources readme (happy to do so).

varyonic:
Merged manually, see https://github.com/activeadmin/inherited_resources/commit/051b3c9e7b22e3d45c090c1c91f3569eafc13d41

PR #391, opened 2015-01-27
Added an example of using begin_of_association_chain with Devise to READ...

tomasmax:
Hi I added to the README an example of using begin_of_association_chain with Devise, because I thinh that it is very common.


rafaelfranca:
Thank you for the pull request but what that section wants to explain is already explained in the previous example. I don't see any need to other example.

PR #397, opened 2015-03-04
use super for permitted_params

stamm:
I have a class with inherit_resources

``` ruby
class Admin::PostsController < Admin::BaseController
  inherit_resources
  …
end
```

And permitted_params defined here in protected

``` ruby
class Admin::BaseController < ActionController::Base
  …
  protected
  …
  def permitted_params
    params.permit!
  end
end
```

It works on inherit_resources version 1.4.1
But after upgrade to 1.6.0 my test had broken.
If define permitted_params in Admin::PostsController in protected or in public — it works
I found way to fix that, and it works.

### DIFF
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -340,6 +340,7 @@ module InheritedResources
       #    end
       #
       def permitted_params
+        return super if defined?(super)
         return nil  unless respond_to?(resource_params_method_name, true)
         {resource_request_name => send(resource_params_method_name)}
       rescue ActionController::ParameterMissing


rafaelfranca:
Could you add a test to it?

stamm:
@rafaelfranca I try to add test at first, but there are mocks everything, not using ActiveRecord who raise exception on real creating =) https://github.com/rails/rails/blob/d424df0a90a91618965c62ea213781f35bf64eec/activemodel/lib/active_model/forbidden_attributes_protection.rb#L21

varyonic:
This seems an odd use case with an available workaround, I'm +1 to close.

javierjulio:
If you are able to add a test for this, we'd be open to it. Let us know, thanks! For now I'll close this.

PR #398, opened 2015-05-01
Changed deprecation notice

stalcottsmith:
This project appears to be maintained by Rafael Mendonça França.

I believe it is a worthy project and the deprecation notice turns people away from using it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,3 @@
-## Deprecation notice
- 
-Since Rails 3 came out, I have no longer used Inherited Resources and it is no
-longer actively maintained. I have found that the responders abstraction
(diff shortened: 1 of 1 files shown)


stalcottsmith:
This is actually more of a question.  I used this very effectively some years ago and I am not sure if Rails has changed too much for this to be useful anymore.  What exactly is the status of this project?

rafaelfranca:
It is exactly what the README is saying. It is deprecated a d people should not use it. 

Thank you so much for the question

stalcottsmith:
Fair enough.  I will work on my fork if turns out to be something I want to keep using.  Thanks.

PR #405, opened 2015-12-21
Rails 5 support

marcroberts:


rafaelfranca:
Thank you for the pull request. Inherit Resources is not in active development anymore and will not support Rails 5.

PR #409, opened 2016-03-31
use has_scope 0.7.0

mamantoha:


rafaelfranca:
Thank you for the pull request but this gem is deprecated and not in active development.

mamantoha:
Ok. I will use my branch :)

PR #410, opened 2016-06-22
Use has_scope 0.7.0

lambda2:


rafaelfranca:
Thank you for the pull request. Like explained in the README this gem is deprecated and it is not accepting new features.

PR #411, opened 2016-07-01
Version bumps for Rails 5

timoschilling:
I know that inherited_resources is deprecated and not maintained, but please @rafaelfranca help as to keep @activeadmin alive. We depending on inherited_resources. ActiveAdmin works with Rails 5 and this inherited_resources change.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,9 @@
 language: ruby
 rvm:
-  - 1.9.3
-  - 2.0.0
-  - 2.1.5
-  - 2.2.0
-  - jruby-19mode # JRuby in 1.9 mode
+  - 2.1.9
+  - 2.2.5
+  - 2.3.1
+  - jruby-9.0.5.0
   - rbx-2.2.10
 gemfile:
   - test/gemfiles/Gemfile-Rails-3-2
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -1,3 +1,9 @@
+# Version 1.7.0
+
+* Support Rails 5
+* Remove support for Ruby < 2.1
+* Fix URL helpers on mountable engines
+
 # Version 1.6.0
 
 * Support Rails 4.2.
--- a/inherited_resources.gemspec
+++ b/inherited_resources.gemspec
@@ -17,6 +17,8 @@ Gem::Specification.new do |s|
   s.files         = Dir["app/**/*", "lib/**/*", "README.md", "MIT-LICENSE"]
   s.require_paths = ["lib"]
 
+  s.required_ruby_version = '>= 2.1'
+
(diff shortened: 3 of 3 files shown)


rafaelfranca:
Hey thank you for the pull request. I can do it but can I help you in a better way, removing the inherit_resource dependency? How do you feel about this? Do you think it would be a lot of work?

timoschilling:
Thanks for the offer, but I have tried a year ago and it wasn't easy. We plan it for our 2.0 release. And yes you could help us.

For our 1.x it would be great if inherited_resources can support Rails 5.

rafaelfranca:
Definitely I don't want to block you, so I'm fine with merging and releasing this. Could you look at the tests failures? If not it is fine, I can find some time to look on then.

matthewrudy:
A bunch of the tests fail because `has_scope [0.7] requires Ruby version >= 2.1.7.`

What's the best way to resolve that,
set all the gemfiles to use 0.6?
Check the ruby version and switch between 0.6 and 0.7?

timoschilling:
I'm working on a matrix, first try: https://travis-ci.org/activeadmin/has_scope/builds/141531656

@rafaelfranca it is posible to build a special has_scope version, wich is between 0.6 and 0.7?

rafaelfranca:
I'd go with removing support to ruby < 2.1 for the next release.

huerlisi:
I just upgraded my app to Rails 5 and am using this inherited_resources changes. Works like a charm!

I'd also go with a version that removes support for ruby < 2.1.

seanlinsley:
I just added a commit to require Ruby >= 2.1. I also updated the changelog.

rafaelfranca:
It is still broken with Rails 3.2 but I can fix it later.

wvteijlingen:
Thanks for this merge! Would it be possible to release a new version on Rubygems so that we don't have to rely on the git feature of Bundler?

alfmatos:
This breaks the Scaffold generator:

```
Could not find "api_controller.rb" in any of your source paths. Your current source paths are:
/Users/alfmatos/.rvm/gems/ruby-2.3.1/bundler/gems/inherited_resources-4434f0ae72f7/lib/generators/rails/templates
```

Any ideas?

alfmatos:
This solves it in application.rb, as per https://github.com/josevalim/inherited_resources/issues/195:

```
config.app_generators.scaffold_controller = :scaffold_controller
```

mrcasals:
Hi @rafaelfranca! This PR addeda `Version 1.7.0` entry in the Changelog file, but this version is not released in Rubygems. Could you release this version, please? I'll use the code from master meanwhile.

Thank you!

PR #412, opened 2016-07-06
Update gemspec to allow use with rails-5

mattclar:


rafaelfranca:
Thank you for the pull request. We already have one to do the same thing. I'm in the mobile and can't link but you can find it easily in the pull requests tab

Issue #417, opened 2016-08-24
loading issue with Rails 5 in some cases


timoschilling:
There happens a strange error when Rails 5 is in use and a engine is used and a namespace which has no engine is used.

All this happens if you use Rails5, ActiveAdmin and the Knock or MailyHerald::Webui gem together.

Here is a App to reproduce the error: https://github.com/brentmulligan/aa-knock-issue
Stacktrace for the error: https://gist.github.com/brentmulligan/ad2ff32b27c3a20b0c9db8e5acd90ca7
(THX to @brentmulligan)

After some debugging I found out that this rescue don't capture the LoadError.

I have tried to fix it in #416, but it brings new problems.

related to: activeadmin/activeadmin#4532

Frexuz:
Any progress on this (for activeadmin/activeadmin#4532)? Still stopping us from going live :/
The fix in #418 works, but since it's not merged, I'm not sure if its safe to use?

PR #418, opened 2016-10-11
Fix rescue from LoadError in engine loading

OlegSmelov:
Fix for #417.

The same exact idea as in #416, but with a different set of exception types.

cc @timoschilling @brentmulligan


OlegSmelov:
PR build fails the same way and for the same Ruby and Rails versions as [the last master build](https://travis-ci.org/josevalim/inherited_resources/builds/145314410), so I guess I didn't break anything.

rafaelfranca:
Can we try to add a test for it? The fix looks good though

OlegSmelov:
Updated.

PR #424, opened 2017-02-07
Add has_scope bump in the CHANGELOG

rafaelfranca:
Also can we disable the protected branch in master? It is not good to not being able to do small commits like this directly.

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -3,6 +3,8 @@
 * Support Rails 5
 * Remove support for Ruby < 2.1
 * Fix URL helpers on mountable engines
+* Allow support to has_scope > 0.6 and < 1.0. Users are now able to choose which
+  version they want to use in their applications.
 
 # Version 1.6.0
 


CHANGELOG
```diff
@@ -3,6 +3,8 @@
 * Support Rails 5
 * Remove support for Ruby < 2.1
 * Fix URL helpers on mountable engines
+* Allow support to has_scope > 0.6 and < 1.0. Users are not able to choose which
+  version they want to use in their applications.
```
tf: I think, it should be "now able", right?

deivid-rodriguez: Do you mean "are able"?

rafaelfranca: :+1:

deivid-rodriguez:
@rafaelfranca I think the way @timoschilling configures this is to allow admins to override the setting. In any case, I find it nice so people get a chance to review even the small changes. In this case, CHANGELOG could've become more confusing for a while if this had landed straight into master.

rafaelfranca:
My experience working in this project for more than 6 years and in OSS in general is that restricting contributors to do their work without peer review decrease the number of contributions if they are not motivated to do it. In my case I'm not motivated to work in this project but I'd occasionally help, but restricting what I can do will just make me avoid contributing.

I do agree it is nice to give people opportunity to review, but review can happen in different stages. I personally review after the fact to unblock others.

varyonic (approved):

varyonic:
Sorry Rafael, it's just the way we ActiveAdmin maintainers have agreed to work together for now.  As David noted we believe you can override if needed.

rafaelfranca:
I can't. I'm not an admin.

rafaelfranca:
BTW, I'm still a contributor of this project. I did agree to move the project to activeadmin org, but I didn't agree to remove me as contributor neither to change the way the project works.

timoschilling:
@rafaelfranca That automatically happens during the movement from jose to activeadmin. I have invited you now.

rafaelfranca:
Thanks! Thank you for giving a new home to this project and helping with it.

timoschilling:
Your welcome, we have to thank you for working on inherited_resources for rails 4 and rails 5 ❤️

PR #426, opened 2017-02-07
Refactoring engine and initialization

rafaelfranca:
Move the engine to its own file and use use `ActiveSupport.on_load` hooks to extend `ActionController::Base`.

This will avoid loading `ActionController::Base` at boot time and mess with the Rails initialization process.


lib/inherited_resources.rb
```diff
@@ -21,18 +21,12 @@ module InheritedResources
...
-    if config.respond_to?(:app_generators)
-      config.app_generators.scaffold_controller = :inherited_resources_controller
-    else
-      config.generators.scaffold_controller = :inherited_resources_controller
-    end
-  end
 end
 
-class ActionController::Base
+ActiveSupport.on_load(:action_controller) do
```
rafaelfranca, line 28: We could move this to the engine and I think it would be better but I'm not sure if people uses inherit_resources without Rails. Do you think it is fine to us to move it there?

timoschilling: I don't believe that some use inherit_resources without Rails. I think it's fine.

rafaelfranca: Changed the code to be in the engine.

lib/inherited_resources/engine.rb
```diff
@@ -0,0 +1,29 @@
...
+
+    initializer "include inherited_resources in action controller" do
+      ActiveSupport.on_load(:action_controller) do
+        # We can remove this check and change to `on_load(:action_controller_base)` in Rails 5.2.
+        return unless is_a?(ActionController::Base)
+
+        # If you cannot inherit from InheritedResources::Base you can call
+        # inherit_resources in your controller to have all the required modules and
+        # funcionality included.
+        def self.inherit_resources
```
rafaelfranca: Do you think it is worth to put this method in a module and use `extend` here? I'm fine with both bur using a module seems cleaner.

timoschilling: No it's ok

rafaelfranca:
🙈 I think I broke the build. Let me try to fix

lib/inherited_resources.rb
rafaelfranca: We had to keep this here since we need this behavior in our tests and the railtie is not executed in our tests. I can change it to be a method that include this behavior but I prefer to do in a different PR.

deivid-rodriguez:
@rafaelfranca This looks great to me, although I'm not an engines expert! :)

The only thing I don't quite understand is this part of the second commit message: "Also move the initialization to an engine". Also, would it be better to call the engine class `Engine` instead of `Railtie`? And if that's a good change, could that break compatibility in any way?

rafaelfranca:
> The only thing I don't quite understand is this part of the second commit message: "Also move the initialization to an engine"

Good point. I moved the initializer to the engine but later I reverted and forgot to remove the text from the commit message. Thanks.

> Also, would I be better to call the engine class Engine instead of Railtie? And if that's a good change, could that break compatibility in any way?

I though about it too. Usually people doesn't reference gem's engines/railties but to avoid a possible breaking change I kept the name because an engine is a railtie so the name is not incorrect, it is only not specific enough.

deivid-rodriguez:
Ah, I see. That was the potentially breaking change for non-rails users that is not finally happening in this PR, right?

rafaelfranca:
> That was the potentially breaking change for non-rails users that is not finally happening in this PR, right?

In theory it was a breaking change but I don't think users of this gem use it outside a rails app. The main reason I leave it out of this PR is that I'd have to change our test suite because it is a client of this gem that doesn't execute the railtie. Since using `on_load` already fixed the problem I wanted to fix I kept the move to the railtie out of this.

deivid-rodriguez (approved):
Looks good to me!

PR #430, opened 2017-03-07
Fix deprecation of interpolation_options in responders

noahpatterson:
The `interpolation_options` method is deprecated. Update to the suggested `flash_interpolation_options` method. This is only a documentation change to the suggested override.


varyonic (approved):
I see this was changed in responders 2.3.

Fivell (approved):
reasonable for me, https://github.com/plataformatec/responders/blob/master/CHANGELOG.md#230

PR #437, opened 2017-06-06
Test support for Ruby 2.4, drop unsupported configurations

varyonic:
Add Ruby 2.4, released December 2016.
Drop Ruby 2.1, de-supported April 2017.
Community support for Rails 3.2/4.0 ended December 2015, 4.1 in 2016.

### DIFF
--- a/inherited_resources.gemspec
+++ b/inherited_resources.gemspec
@@ -17,10 +17,10 @@ Gem::Specification.new do |s|
   s.files         = Dir["app/**/*", "lib/**/*", "README.md", "MIT-LICENSE"]
   s.require_paths = ["lib"]
 
-  s.required_ruby_version = '>= 2.1'
+  s.required_ruby_version = '>= 2.2'
 
   s.add_dependency("responders")
-  s.add_dependency("actionpack", ">= 3.2", "< 5.2.x")
-  s.add_dependency("railties", ">= 3.2", "< 5.2.x")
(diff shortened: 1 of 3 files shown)


deivid-rodriguez:
I think if we officially drop support, we should bump the minimum versions in gemspec. And if we keep it, we should keep testing it.

I'm up for dropping anything under ruby 2.2 and rails 4.2.

rafaelfranca:
Good idea! But we need to change the gemspec indeed.

varyonic:
Let's leave this open until we have a better reason to bump and release.

inherited_resources.gemspec
```diff
@@ -17,7 +17,7 @@ Gem::Specification.new do |s|
   s.files         = Dir["app/**/*", "lib/**/*", "README.md", "MIT-LICENSE"]
   s.require_paths = ["lib"]
 
-  s.required_ruby_version = '>= 2.1'
+  s.required_ruby_version = '>= 2.2'
 
   s.add_dependency("responders")
   s.add_dependency("actionpack", ">= 3.2", "< 5.2.x")
```
rafaelfranca: I think we need to bump this too since now we don't test with 3.2, 4.0 and 4.1

varyonic:
Corrected gemspec and updated CHANGELOG.  I propose a release 1.8 accompanies activeadmin 1.1 when ready.

rafaelfranca (approved):
This have some breaking changes so I'd call it 2.0 instead of 1.8.

PR #438, opened 2017-06-14
Please add support for API controllers.

luismasuelli:
Please add support to API controllers as well!

When I had to use API controllers with IR, I found that `inherit_resources` method was not available and found the source code was like this. Instead, I had to call the three lines in the method body. It would be great if the support was somehow built-in or at least documented. The API controllers worked exactly as I expected.

### DIFF
--- a/lib/inherited_resources.rb
+++ b/lib/inherited_resources.rb
@@ -27,7 +27,7 @@ end
 
 ActiveSupport.on_load(:action_controller) do
   # We can remove this check and change to `on_load(:action_controller_base)` in Rails 5.2.
-  if self == ActionController::Base
+  if self == ActionController::Base || self == ActionController::API
     # If you cannot inherit from InheritedResources::Base you can call
     # inherit_resources in your controller to have all the required modules and
     # funcionality included.


lib/inherited_resources.rb
```diff
@@ -27,7 +27,7 @@ def self.flash_keys=(array)
 
 ActiveSupport.on_load(:action_controller) do
   # We can remove this check and change to `on_load(:action_controller_base)` in Rails 5.2.
-  if self == ActionController::Base
+  if self == ActionController::Base || self == ActionController::API
```
rafaelfranca, line 30: Just remove this conditional is you want to support API controllers along with the comment above.

rafaelfranca (changes requested):
I'm not sure if we want to support API controllers. Inherit Resources was made to navigational controllers, the new and edit actions for example doesn't make a lot of sense in APIs, but I'm not totally against.

luismasuelli:
Well, even if my solution is not to be merged as-is (actually it was not my intention an instant merge, but an analysis of the points and a sample attention to the issue), but just an illustrative one.

However, I found quite useful my "hack" for my API controller (copying and directly using source lines inside `inherit_resources` in my custom API controller, using the currently deployed/distributed version of IR - not the line I suggested since that involves a change in IR) and in a near future I'll contribute with a better-suited change in this PR (please keep it open a couple of days until I come with a better suggestion I have in mind right now).

________________________________
De: Rafael França <notifications@github.com>
Enviado: jueves, 15 de junio de 2017 01:13:03 a.m.
Para: activeadmin/inherited_resources
Cc: luismasuelli; Author
Asunto: Re: [activeadmin/inherited_resources] Please add support for API controllers. (#438)


@rafaelfranca requested changes on this pull request.

I'm not sure if we want to support API controllers. Inherit Resources was made to navigational controllers, the new and edit actions for example doesn't make a lot of sense in APIs, but I'm not totally against.

________________________________

In lib/inherited_resources.rb<https://github.com/activeadmin/inherited_resources/pull/438#discussion_r122116253>:

> @@ -27,7 +27,7 @@ def self.flash_keys=(array)

 ActiveSupport.on_load(:action_controller) do
   # We can remove this check and change to `on_load(:action_controller_base)` in Rails 5.2.
-  if self == ActionController::Base
+  if self == ActionController::Base || self == ActionController::API


Just remove this conditional is you want to support API controllers along with the comment above.

varyonic:
I'm +1 for declining this new functionality as noted in the README, others can always fork if they find it useful.

rafaelfranca:
Right. Closing. Thank you so much for the pull request

PR #440, opened 2017-11-11
Fix URL helpers not working with ActionController::Parameters

### LINKED ISSUE #439: Parameters are ignored in Rails 5.1
I'm upgrading an old Rails 3.2 application to 5.1. This application uses inherited resources and I noticed that the following doesn't work:

```
collection_url(params.permit!.merge(example: true))
```

while this one works:

```
users_url(params.permit!.merge(example: true))
```

The issue seems to come from extract_options! in https://github.com/activeadmin/inherited_resources/blob/5d18499877852437f2c058747f07bc80e46557fc/lib/inherited_resources/url_helpers.rb#L223 and https://github.com/activeadmin/inherited_resources/blob/5d18499877852437f2c058747f07bc80e46557fc/lib/inherited_resources/url_helpers.rb#L229

Previously something like params.permit!.merge(example: true) would return a Hash and now it returns an ActionController::Parameters, which won't work with extract_options!.

I was able to fix this by using something like:

```
          def #{prefix}#{name}_url(*given_args)
            given_options = given_args.collect { |arg| arg.is_a?(ActionController::Parameters) ? arg.to_h : arg }.extract_options!
            #{prefix}#{segments}_url(#{ivars})
          end
```

If this change is correct, I can prepare a PR.

fernandomm:
Helpers like collection_url/path use extract_options! which won't work with ActionController::Parameters since it isn't a hash.

This PR adds a check and transform ActionController::Parameters to a hash if necessary, making URL helpers behave like the Rails ones when parameter is a ActionController::Parameters object.

Fix #439.

### DIFF
--- a/lib/inherited_resources/url_helpers.rb
+++ b/lib/inherited_resources/url_helpers.rb
@@ -216,7 +216,7 @@ module InheritedResources
       prefix = prefix ? "#{prefix}_" : ''
       ivars << (ivars.empty? ? 'given_options' : ', given_options')
 
-      given_args_transform = defined?(ActionController::Parameters) ? 'given_args = given_args.collect { |arg| arg.respond_to?(:permitted?) ? arg.to_h : arg }' : ''
+      given_args_transform = 'given_args = given_args.collect { |arg| arg.respond_to?(:permitted?) ? arg.to_h : arg }'
 
       class_eval <<-URL_HELPERS, __FILE__, __LINE__
         protected
--- a/lib/inherited_resources/url_helpers.rb
+++ b/lib/inherited_resources/url_helpers.rb
@@ -216,7 +216,7 @@ module InheritedResources
       prefix = prefix ? "#{prefix}_" : ''
       ivars << (ivars.empty? ? 'given_options' : ', given_options')
 
-      given_args_transform = defined?(ActionController::Parameters) ? 'given_args = given_args.collect { |arg| arg.is_a?(ActionController::Parameters) ? arg.to_h : arg }' : ''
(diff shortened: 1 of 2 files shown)


lib/inherited_resources/url_helpers.rb
```diff
@@ -216,16 +216,20 @@ def generate_url_and_path_helpers(prefix, name, resource_segments, resource_ivar
       prefix = prefix ? "#{prefix}_" : ''
       ivars << (ivars.empty? ? 'given_options' : ', given_options')
 
+      given_args_transform = defined?(ActionController::Parameters) ? 'given_args = given_args.collect { |arg| arg.is_a?(ActionController::Parameters) ? arg.to_h : arg }' : ''
```
rafaelfranca: `'given_args = given_args.collect { |arg| arg.respond_to?(:permitted?) ? arg.to_h : arg }'`

rafaelfranca (changes requested):

lib/inherited_resources/url_helpers.rb
```diff
@@ -216,16 +216,20 @@ def generate_url_and_path_helpers(prefix, name, resource_segments, resource_ivar
       prefix = prefix ? "#{prefix}_" : ''
       ivars << (ivars.empty? ? 'given_options' : ', given_options')
 
+      given_args_transform = defined?(ActionController::Parameters) ? 'given_args = given_args.collect { |arg| arg.respond_to?(:permitted?) ? arg.to_h : arg }' : ''
```
rafaelfranca: You don't need to check if `ActionController::Parameters` is defined anymore with the actual code.

rafaelfranca (changes requested):

rafaelfranca (approved):

rafaelfranca:
Thanks!

PR #443, opened 2017-11-30
Verify support for Rails 5.2.

varyonic:


lib/inherited_resources/version.rb
```diff
@@ -1,3 +1,3 @@
 module InheritedResources
-  VERSION = '1.7.2'.freeze
+  VERSION = '2.0.0'.freeze
```
nunosilva800: historically, support for newer versions of rails was done as a patch update (1.7.1 -> 1.7.2 for example).
see https://github.com/activeadmin/inherited_resources/blob/master/CHANGELOG

Perhaps this should be 1.7.4, or 1.8.

varyonic: Agreed, but I'll defer to @rafaelfranca https://github.com/activeadmin/inherited_resources/pull/437#issuecomment-308746524

rafaelfranca: `1.8.0` is good for me since there is no breaking change on this release.

rafaelfranca (approved):
I'd just bump to 1.8.0. Otherwise looks good.

PR #446, opened 2018-02-21
[CI] Test against Ruby 2.x

nicolasleger:


varyonic:
Do we really need to test *every* version of MRI? Would min (2.2) and max (2.5) be sufficient?

rafaelfranca:
Maybe, but maybe not. We could write code that works in 2.2 and 2.5 but not in 2.4 for example.

rafaelfranca (approved):

PR #449, opened 2018-07-31
Fix dependencies so gem works with rails 5.2.1.rc1 and all other 5.2.x versions

jpawlyn:
This resolves a bundle update problem for rails 5.2.1.rc1. The relevant output of `bundle update rails` is:

```
Bundler could not find compatible versions for gem "actionpack":
  In Gemfile:
    activeadmin (~> 1.3) was resolved to 1.3.0, which depends on
      inherited_resources (>= 1.7.0) was resolved to 1.8.0, which depends on
        actionpack (<= 5.2, >= 4.2)
```

### DIFF
--- a/inherited_resources.gemspec
+++ b/inherited_resources.gemspec
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.2'
 
   s.add_dependency("responders")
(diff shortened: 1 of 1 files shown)


rafaelfranca (approved):

jpawlyn:
Please note there's an additional issue with the polyamorous gem - see https://github.com/activerecord-hackery/polyamorous/issues/41

PR #453, opened 2018-09-18
Tidelift support

varyonic:


README.md
```diff
@@ -11,6 +11,7 @@ If you are not already using Inherited Resources we suggest instead using Rails'
 
 [![Version](http://img.shields.io/gem/v/inherited_resources.svg)](https://rubygems.org/gems/inherited_resources)
 [![Travis CI](http://img.shields.io/travis/activeadmin/inherited_resources/master.svg)](https://travis-ci.org/activeadmin/inherited_resources)
+[![Tidelift        ][tidelift_badge]][tidelift]
```
deivid-rodriguez: I would either use the same style as existing badges, or extract all bagde links to the bottom.

deivid-rodriguez (approved):

rafaelfranca:
Why the gemfile.lock was removed? This is an antipattern.

rafaelfranca:
https://bundler.io/blog/2018/01/17/making-gem-development-a-little-better.html

deivid-rodriguez:
@rafaelfranca I agree with you (it's actually me being mentioned in that blog post :smiley:) but I never managed to convince @varyonic, @timoschilling and the other maintainers... Since the rest of the gemfiles had already been removed in another PR and I didn't want to discuss the same thing again, I approved it.

But I'm definitely open to reconsider if we manage to convince the rest!

rafaelfranca:
I think they should convince us that it should be removed, not the other way around. This gem has the Gemfile.lock for a long time ago so if we want to remove we should give good argument to why we should remove it, not the opposite. If I was going to add the Gemfile.lock I agree I should need to convince people to do so.

javierjulio:
For the record haha as one of the other maintainers I've always been convinced to keep it. I agree with @rafaelfranca points.

deivid-rodriguez:
Yeah, sorry @javierjulio, "and the other maintainers" was not accurate. I think only @varyonic and @timoschilling were opposed.

Anyways, I was not paying any attention to this repo when the gemfiles were removed and that PR was approved and went in, so now I feel I need to convince @varyonic (and not the other way around) if I want to revert the change :sweat_smile:. And I decided to approve this PR and leave that for another time.

javierjulio:
@deivid-rodriguez no need to apologize! Its fine and easy to correct. I just wanted you and @rafaelfranca to know I support having the Gemfile.lock. 👍🏻 I remember that you worked on this for bundler and thought it was a great improvement! I know since I've had to ask you for help when doing ActiveAdmin dev and it had to do with dependencies being out of sync. 😂

deivid-rodriguez:
I'll propose a workflow to try to fit everyone's taste here:

* Re-add lock files to the repo. @rafaelfranca will be happy, I'll be happy, @javierjulio will be happy.

* Remove the lockfiles in CI before running tests. I think this will make @timoschilling happy because his objection was that bug discoverability on latest dependencies would be reduced. Also @rafaelfranca will be happy since I think this technique is used in Rails. Also @varyonic will be happy because one of his objections was that it was too much work to keep them up to date. By testing against latest versions that work will be nice to do but less necessary.

* Update lockfiles not one dependency at a time nor with a bot, but with `bundle update --all`. I think this will also make @varyonic happy, because another of his objections was that keeping the lockfiles up to date would be a lot of git noise, and with this approach the "noise" should be reduced. It will also make @rafaelfranca happy because it's what's done in Rails too I think.

If you like my proposal I can make a PR!

rafaelfranca:
👍 for this proposal

deivid-rodriguez:
Ok, I'll wait for @varyonic's input and then make the PR. I think he should be ok with this!

varyonic:
@rafaelfranca To answer your question as far as I could see this lockfile was not used since 6d034d913bfd9b53ea2e5b31c4056578a1b170a4  @deivid-rodriguez If this is good enough for Rails then it's good enough for Active Admin... all three repos please.

deivid-rodriguez:
Actually I misunderstood e258e88. It was removing Gemfiles, not lock files. So now this repo does not have a `Gemfile` in the root folder, so running `bundle` from there will fail... right?

I'm not sure if this is a problem at all, but I guess some people are too used to running `bundle install` when cloning a repo so that might surprise them. I think we could put the latest `Gemfile` in the root folder, so that things work against that by default? Or maybe symlink it? Or maybe tell contributors to do `export BUNDLE_GEMFILE=test/gemfiles/Gemfile-5-2`? I think that last one is what we do in activeadmin.

I think I'd go with the first option.

rafaelfranca:
It is used when you clone the repository and that is what is used when contributors clone this repository and run tests. Without that lockfile contributors may close this repository and get a red build because one of the dependencies is incompatible with the current code, and that is confusing to them since they don't know they should ignore that or if they should fix it. With that lockfile that problem goes away because we know the gem versions that are going to be installed on their machines are the same we used in our machines and we know have a green build.

deivid-rodriguez:
Yeah, actually not only that but it can also happen that they cloned the repo a long time ago, go back to work on it after a while, pull the latest `Gemfile` and it's so different from the `Gemfile.lock` they have in their local folder that bundler freaks out. This is no big deal for experienced users, probably, but it can discourage beginners.

Anyways now I was not talking about lockfiles not being in the repo, but about the repo not having a `Gemfile`. Right now users cloning the repo and running `bundle` on it will get  a "Could not locate Gemfile" error. In my opinion the easiest fix for that is to move `test/gemfiles/Gemfle-5-2` to the root of the repo.

PR #454, opened 2018-09-19
Gemfile management

### LINKED ISSUE #445: Bump nokogiri dependency
Gemnasium [flagged this project](https://gemnasium.com/github.com/activeadmin/inherited_resources/alerts) as a security alert was fixed for nokogiri (libxml) in 1.8.2.  That's the downside of having Gemfile.lock included in the repo.

deivid-rodriguez:
This is my proposal after the discussion in #453.

Closes #445.


deivid-rodriguez:
I should say that I'm not a fan of the `rm ${BUNDLE_GEMFILE}.lock` trick but I wanted to find the middle ground, so here it is!

.travis.yml
```diff
@@ -5,11 +5,16 @@ rvm:
...
+
 gemfile:
   - test/gemfiles/Gemfile-Rails-4-2
   - test/gemfiles/Gemfile-Rails-5-0
   - test/gemfiles/Gemfile-Rails-5-1
-  - test/gemfiles/Gemfile-Rails-5-2
+  - Gemfile
+
+script:
+  - rm "${BUNDLE_GEMFILE}.lock" && bundle exec rake
```
rafaelfranca: I think the rm should be before bundle runs. So a `before_install` hook like this one? https://github.com/rails/rails/blob/master/.travis.yml#L34

deivid-rodriguez: Oooops, you're right! I'm surprised this solution didn't actually fail!

deivid-rodriguez:
It should be fixed now!

rafaelfranca (approved):

javierjulio (approved):

javierjulio:
@deivid-rodriguez thanks for doing this for all the repos! And thanks to you @rafaelfranca for helping us straighten this out. Much appreciated! 🙏🏻❤️

PR #455, opened 2018-10-05
Refactor url helpers

varyonic:
The current implementation of UrlHelpers#create_resources_url_helpers! uses a class eval of something like:
```ruby
  edit_book_folder_path(@book,(given_args.first || @folder),given_options)
```
This refactoring replaces the class eval with a define_method as [suggested by Aaron](https://tenderlovemaking.com/2013/03/03/dynamic_method_definitions.html).  The proposed implementation looks something like:
```ruby
  edit_book_folder_path _edit_book_folder_params(given_args)
```
The presupposition is that using a defined method is easier to reason about and troubleshoot than eval-ing a constructed list of interpolated parameters.

This module could possibly be clarified further but the scope of this PR is only the refactoring required to replace use of class_eval.  A future refactoring could, for example, decompose the params helper further for different scenarios.

### DIFF
--- a/lib/inherited_resources/url_helpers.rb
+++ b/lib/inherited_resources/url_helpers.rb
@@ -223,7 +223,7 @@ module InheritedResources
 
       undef_method params_method_name if method_defined? params_method_name
 
-      define_method params_method_name do |given_options, *args|
+      define_method params_method_name do |given_args, given_options, *args|
         args << given_options
       end
       protected params_method_name
@@ -240,7 +240,7 @@ module InheritedResources
         def #{method_name}(*given_args)
           given_args = given_args.collect { |arg| arg.respond_to?(:permitted?) ? arg.to_h : arg }
           given_options = given_args.extract_options!
-          #{segments_method}(*#{params_method_name}(given_options, #{ivars.join(?,)}))
+          #{segments_method}(*#{params_method_name}(given_args, given_options, #{ivars.join(?,)}))
         end
       URL_HELPERS
       protected method_name
--- a/lib/inherited_resources/url_helpers.rb
+++ b/lib/inherited_resources/url_helpers.rb
@@ -194,36 +194,31 @@ module InheritedResources
       # In collection in polymorphic cases, allow an argument to be given as a
       # replacemente for the parent.
       #
-      if name == :collection && polymorphic
-        index = ivars.index(:parent)
-        ivars.insert index, "(given_args.first || parent)"
-        ivars.delete(:parent)
-      end
+      parent_index = ivars.index(:parent) if name == :collection && polymorphic
 
-      # When polymorphic is true, the segments must be replace by :polymorphic
-      # and ivars should be gathered into an array, which is compacted when
-      # optional.
-      #
-      if polymorphic
-        segments = :polymorphic
-        ivars    = "[#{ivars.join(', ')}]"
-        ivars   << '.compact' if self.resources_configuration[:polymorphic][:optional]
+      segments = if polymorphic
+        :polymorphic
+      elsif resource_segments.empty?
+        'root'
       else
-        segments = resource_segments.empty? ? 'root' : resource_segments.join('_')
+        resource_segments.join('_')
       end
 
       ivars = ivars.present? ? Array(ivars) : []
 
-      define_params_helper(prefix, name, ivars)
+      define_params_helper(prefix, name, polymorphic, parent_index)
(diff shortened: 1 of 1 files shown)


lib/inherited_resources/url_helpers.rb
```diff
@@ -184,56 +174,71 @@ def generate_url_and_path_helpers(prefix, name, resource_segments, resource_ivar
...
+        'root'
       else
-        segments = resource_segments.empty? ? 'root' : resource_segments.join('_')
-        ivars    = ivars.join(', ')
+        resource_segments.join('_')
       end
 
-      prefix = prefix ? "#{prefix}_" : ''
-      ivars << (ivars.empty? ? 'given_options' : ', given_options')
+      ivars = ivars.present? ? Array(ivars) : []
```
deivid-rodriguez: After all the refactoring `ivars` is never converted to string, so I think this whole line can just go?

deivid-rodriguez (commented):
It's looking solid to me, nice job!

lib/inherited_resources/url_helpers.rb
```diff
@@ -184,56 +174,71 @@ def generate_url_and_path_helpers(prefix, name, resource_segments, resource_ivar
...
-      class_eval <<-URL_HELPERS, __FILE__, __LINE__
-        protected
-          undef :#{prefix}#{name}_path if method_defined? :#{prefix}#{name}_path
-          def #{prefix}#{name}_path(*given_args)
-            #{given_args_transform}
-            given_options = given_args.extract_options!
-            #{prefix}#{segments}_path(#{ivars})
-          end
+    def define_params_helper(prefix, name, singleton, polymorphic, parent_index, ivars)
+      params_method_name = ['', prefix, name, :params].compact.join(?_)
```
deivid-rodriguez: We've introduced this `?_` new style for single character string literals. I didn't know about that style and I don't think many people do, and the change is not applied consistently anways so... I would keep the regular style.

javierjulio: Wow. I didn't know you could do that! Very cool:
```
>> ['', 'prefix', 'name', :params].compact.join(?_)
=> "_prefix_name_params"
>> ?_
=> "_"
```

Agreed though about keeping the regular style.

lib/inherited_resources/url_helpers.rb
```diff
@@ -184,56 +174,69 @@ def generate_url_and_path_helpers(prefix, name, resource_segments, resource_ivar
...
-          def #{prefix}#{name}_url(*given_args)
-            #{given_args_transform}
-            given_options = given_args.extract_options!
-            #{prefix}#{segments}_url(#{ivars})
+      define_method params_method_name do |*given_args|
+        given_args = given_args.collect { |arg| arg.respond_to?(:permitted?) ? arg.to_h : arg }
+        given_options = given_args.extract_options!
+
+        args = ivars.map do |ivar|
+          ivar.is_a?(Symbol) && ivar.to_s.start_with?(?@) ? instance_variable_get(ivar) : ivar
```
deivid-rodriguez: There's still this leftover `?@` thingy.

lib/inherited_resources/url_helpers.rb
```diff
@@ -184,56 +174,69 @@ def generate_url_and_path_helpers(prefix, name, resource_segments, resource_ivar
...
       else
-        segments = resource_segments.empty? ? 'root' : resource_segments.join('_')
-        ivars    = ivars.join(', ')
+        resource_segments.join('_')
       end
 
-      prefix = prefix ? "#{prefix}_" : ''
-      ivars << (ivars.empty? ? 'given_options' : ', given_options')
+      define_params_helper(prefix, name, singleton, polymorphic, parent_index, ivars)
+      [:path, :url].each { |suffix| define_helper_method(prefix, name, suffix, segments) }
```
rafaelfranca: Maybe instead of this loop we can do the two method calls? This array will not grow over time so this change would remove this indirection.

rafaelfranca (approved):

deivid-rodriguez:
By the way, I tried running this refactor against ActiveAdmin's master, and it all went [well](https://travis-ci.org/activeadmin/activeadmin/builds/439620424)!

deivid-rodriguez:
@varyonic Do you still plan to get this over the line?

deivid-rodriguez (approved):
Looks great to me! :100:

javierjulio (approved):
Great thanks! 👍🏻 I'll get this in.

PR #456, opened 2018-10-08
Build only PRs

deivid-rodriguez:
As with the other repositories, I'd like to save CI time by not having two duplicate TravisCI checks on each PR.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,7 @@
+branches:
+  only:
+    - master
+
 language: ruby
 rvm:
   - 2.3


javierjulio (approved):

varyonic:
WTF? https://github.com/activeadmin/activeadmin/pull/5126#issuecomment-322049874

deivid-rodriguez:
Could you please read the whole thing, starting with https://github.com/activeadmin/activeadmin/pull/5488, and then elaborate a bit more? Thanks.

deivid-rodriguez:
Then you can continue with https://github.com/activeadmin/arbre/pull/79 where I explicitly mention you asking for your opinion.

varyonic:
OK, I give up. I thought this was settled by #5126. I absolutely test all my branches on Travis before submitting PRs.

deivid-rodriguez:
@varyonic Why commenting a "WTF comment" on a merged PR, instead of giving actual reasons when the change is initially proposed (or on a PR [pending review](https://github.com/activeadmin/activeadmin/pull/5491))? You seemed to be trying to point out that I'm contradicting myself, but this is not true, I just accepted the "let's do what Rails does" reason and decided to live with it. When I started this series of PRs, I figured I could propose the change again because Rails no longer does this, so that was no longer a reason against it.

Now I'm doing a deeper git history and I see that you and @timoschilling were [opposed to this change](https://github.com/activeadmin/activeadmin/pull/4811) because you usually test all your branches before opening a PR. If that's still the case, and you don't to go through the trouble of temporarily adding your working branch when you want to get it tested in CI without a PR, I'm fine with reverting this again, I don't want to impose my workflow to everyone. I just don't get destructive late comments over more constructive and early feedback.

javierjulio:
For the record, while I do have a preference for this, I don't want to argue if others feel strongly otherwise. The reason I support this change is that I think its best because:

- it uses less resources
- allows more frequent builds 
- requires both contributors and maintainers to be open by creating PRs here
- feedback is both automated and community based in one place
- consistent behavior/flow for all involved

While it is what I'm used too, this has helped me learn and reason about why I prefer this setup so thanks!

PR #458, opened 2018-10-24
Add changelog test to maintain consistency

javierjulio:
This one doesn't really follow what we do in the ActiveAdmin repo but I think its great and can help us here in maintaining this and other repos over time.

I'm happy to make changes specific to this but for now I focused on maintaining what the test already covers so I had to update the changelog. What's interesting though is it shows a few weak spots. Notice that some list items where a new line is manually added, the test doesn't pick that up. I left that in on purpose to demonstrate and discuss. You'll notice what I mean on line 25 (and others like it).

For reference, this originates from: https://github.com/activeadmin/activeadmin/pull/5474


deivid-rodriguez:
Yeah, this one works for rubocop (where I borrowed these checks from) because each change log entry goes in a single line. So maybe we could enforce that?

Also, the original check in [rubocop](https://github.com/rubocop-hq/rubocop/blob/9f518b6bbbd98857c19b64f036e64bdd2ccf255e/spec/project_spec.rb#L167) actually enforces punctuation at the end of entries. For activeadmin I removed it because we also include a mention to contributor and PR after the punctuation, so our entries ended _without_ a dot. So, maybe we actually enforce a final dot here? I think it looks better and it also matches our current style.

Thoughts?

deivid-rodriguez:
Thanks for doing this by the way!

javierjulio:
@deivid-rodriguez yes that sounds good to me! I'll circle back and get those changes in. The one addition we'd have to do here though is figure out how to enforce entries to be on a single line?

deivid-rodriguez:
Yes, I guess we could enforce that all lines are either blank lines or start with `#` or `*`?

rafaelfranca:
If we are going to enforce something I prefer to each line to end with punctuation.

javierjulio:
@rafaelfranca yes I will do that. @deivid-rodriguez pointed that out too and I'm good with that. 👍🏻 I was just going off what we had in ActiveAdmin as a starting point since I wasn't entirely sure which way we'd prefer.

rafaelfranca:
I'm fine with a single entry per line to make if that is easier to enforce too.

deivid-rodriguez:
By the way, I'm working on a tool to automatically add changelog entries right after PRs are merged into master, by parsing the PR title and all the rest of the information from the merged PR. So we get a consistent changelog format, and we don't need to bother contributors with changelog entries! :smiley:

javierjulio:
@deivid-rodriguez unless I'm mistaken I realized the test could stay simple. Any list item (line starting with a *) line now has to end with either a "." or ":" so I didn't think how I could include newlines there. In a way they are already taken care of since it won't allow it. Lines that had hard returns I put onto a single line. Do you think that is sufficient for now or you had other ideas for what we can do here?

deivid-rodriguez (approved):

javierjulio:
@deivid-rodriguez great thanks! Let me know as I'm happy to iterate on this too if you have other ideas. I plan to do the same for arbre as well as I have time.

PR #459, opened 2018-10-26
Fix warnings from running the test suite

javierjulio:
When running the test suite, many warnings are being displayed which we should fix. To see the warnings alongside test case names, run tests using the verbose flag:

```
TESTOPTS=--verbose bundle exec rake
```

I'm not sure how easy some warnings will be to address but this is a start to tackle each one. If you have ideas or know how to address any, please feel free to share. Thanks!

## Todo List

- Once explicit resources are known, we can resolve the Rails 6 deprecation warning as we can start localizing required resources into each test rather than have them be global.

### DIFF
--- a/lib/inherited_resources/polymorphic_helpers.rb
+++ b/lib/inherited_resources/polymorphic_helpers.rb
@@ -102,11 +102,13 @@ module InheritedResources
       # as parent types.
       #
       def parent_type
-        unless @parent_type
+        unless instance_variable_defined?(:@parent_type)
           symbols_for_association_chain
         end
 
-        @parent_type
+        if instance_variable_defined?(:@parent_type)
+          @parent_type
+        end
       end
 
       def parent_class
--- a/lib/inherited_resources/polymorphic_helpers.rb
+++ b/lib/inherited_resources/polymorphic_helpers.rb
@@ -120,7 +120,7 @@ module InheritedResources
       #
       def parent
         if parent_type
-          p = instance_variable_get("@#{parent_type}")
+          p = instance_variable_defined?("@#{parent_type}") && instance_variable_get("@#{parent_type}")
           p || instance_variable_set("@#{parent_type}", association_chain[-1])
         end
       end
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -36,6 +36,14 @@ InheritedResources::Routes.draw do
   root to: 'posts#index'
 end
 
+def draw_routes(&block)
+  InheritedResources::Routes.draw(&block)
+end
+
+def clear_routes
+  InheritedResources::Routes.draw { }
+end
+
 ActionController::Base.send :include, InheritedResources::Routes.url_helpers
 
 # Add app base to load path
--- a/test/test_helper.rb
+++ b/test/test_helper.rb
@@ -29,12 +29,6 @@ require 'inherited_resources'
 ActionController::Base.view_paths = File.join(File.dirname(__FILE__), 'views')
 
 InheritedResources::Routes = ActionDispatch::Routing::RouteSet.new
-InheritedResources::Routes.draw do
-  get ':controller(/:action(/:id))'
-  get ':controller(/:action)'
-  resources 'posts'
-  root to: 'posts#index'
-end
 
 def draw_routes(&block)
   InheritedResources::Routes.draw(&block)
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -152,7 +152,10 @@ module InheritedResources
         polymorphic = options.delete(:polymorphic)
         finder      = options.delete(:finder)
 
-        include BelongsToHelpers if self.parents_symbols.empty?
+        if self.parents_symbols.empty?
+          include BelongsToHelpers
+          helper_method :parent, :parent?
+        end
 
         acts_as_polymorphic! if polymorphic || optional
         acts_as_shallow!     if shallow
@@ -218,7 +221,6 @@ module InheritedResources
         else
           create_resources_url_helpers!
         end
-        helper_method :parent, :parent?
       end
       alias :nested_belongs_to :belongs_to
 
--- a/test/aliases_test.rb
+++ b/test/aliases_test.rb
@@ -41,6 +41,16 @@ end
 class AliasesTest < ActionController::TestCase
   tests StudentsController
 
+  def setup
+    draw_routes do
+      resources :students
+    end
+  end
+
+  def teardown
+    clear_routes
+  end
+
   def test_assignments_before_calling_alias
     Student.stubs(:new).returns(mock_student)
     get :new
--- a/test/association_chain_test.rb
+++ b/test/association_chain_test.rb
@@ -31,7 +31,15 @@ class BeginOfAssociationChainTest < ActionController::TestCase
   tests PetsController
 
   def setup
+    draw_routes do
+      resources :pets
+    end
+
     @controller.current_user = mock()
+  end
+
+  def teardown
+    clear_routes
   end
 
   def test_begin_of_association_chain_is_called_on_index
@@ -95,8 +103,16 @@ class AssociationChainTest < ActionController::TestCase
   tests PuppetsController
 
   def setup
+    draw_routes do
+      resources :puppets
+    end
+
     @controller.stubs(:resource_url).returns('/')
     @controller.stubs(:collection_url).returns('/')
+  end
+
+  def teardown
+    clear_routes
   end
 
   def test_parent_is_added_to_association_chain
--- a/test/base_test.rb
+++ b/test/base_test.rb
@@ -26,11 +26,19 @@ end
 
 module UserTestHelper
   def setup
+    draw_routes do
+      resources :users
+    end
+
     @controller_class    = Class.new(UsersController)
     @controller          = @controller_class.new
     @controller.request  = @request = new_request
     @controller.response = @response = new_response
     @controller.stubs(:user_url).returns("/")
+  end
+
+  def teardown
+    clear_routes
   end
 
   protected
--- a/test/belongs_to_test.rb
+++ b/test/belongs_to_test.rb
@@ -56,7 +56,7 @@ class BelongsToTest < ActionController::TestCase
     assert_equal mock_comment, assigns(:comment)
   end
 
-  def test_redirect_to_the_post_on_update_if_show_and_index_undefined
+  def test_redirect_to_the_post_on_create_if_show_and_index_undefined
     @controller.class.send(:actions, :all, except: [:show, :index])
     @controller.expects(:parent_url).returns('http://test.host/')
     Comment.expects(:build).with({'these' => 'params'}).returns(mock_comment(save: true))
--- a/test/belongs_to_test.rb
+++ b/test/belongs_to_test.rb
@@ -15,6 +15,7 @@ end
 class BelongsToTest < ActionController::TestCase
   tests CommentsController
 
+  # TODO: look into the failures as this should run randomly
   def self.test_order
     # version 5 defaults to random, which fails...
     MiniTest::Unit::VERSION.to_i >= 5 ? :alpha : super
--- a/test/belongs_to_test.rb
+++ b/test/belongs_to_test.rb
@@ -21,11 +21,19 @@ class BelongsToTest < ActionController::TestCase
   end
 
   def setup
+    draw_routes do
+      resources :comments, :posts
+    end
+
     Post.expects(:find).with('37').returns(mock_post)
     mock_post.expects(:comments).returns(Comment)
 
     @controller.stubs(:resource_url).returns('/')
     @controller.stubs(:collection_url).returns('/')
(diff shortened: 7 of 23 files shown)


javierjulio:
Hmm it seems some test cases have the same name, for example, in `nested_belongs_to_test.rb` there are 2 cases with the same name of `test_assigns_country_and_state_and_city_on_create` but only one uses the create action (the second, latter instance). Some I renamed to be unique but after awhile some started to cause a failure. I wasn't sure why at the time but now its clear. The test case (method) really is being discarded, meaning its not even running that test case! Once I rename the test case to be unique I see a change in the number of test cases that run which confirm that. These will be fixed.

javierjulio:
Ok after this we are left with just these warnings:

* DEPRECATION WARNING: Using a dynamic :controller segment in a route is deprecated and will be removed in Rails 6.0. (test_helper.rb:46 and 47)
* For the `parent` method: 
  ```
  helpers.rb:66: warning: method redefined; discarding old parent
  helpers.rb:66: warning: previous definition of parent was here
  ```
* For. the `parent?` method:
  ```
  helpers.rb:66: warning: method redefined; discarding old parent?
  helpers.rb:66: warning: previous definition of parent? was here
  ```

The Rails deprecation warning I know how to resolve but I think it will take awhile because the tests are so reliant on requests. I don't know what's wrong with the `parent/?` methods though.

deivid-rodriguez:
@javierjulio Thanks so much for this work! It's great to have a warningless test suite! :muscle:

I'm on a trip so I'm not sure I'l be able to review until Tuesday, but regarding the `parent?` warnings, those kind of warning will happen, for example, if you `load` (not `require`, `load`) the same file twice. Not sure if that gives you any clues.

deivid-rodriguez:
I had a quick look. I think the problem is that the same test class is reused in several tests, calling `belongs_to` on it several times (see for example usages of `Library::SubcategoriesController`). Each of these calls redefines these methods. Possible solutions would be to undefine them after each test, or to use different classes for each test.

deivid-rodriguez:
Also, regarding the addition of `minitest-reporters`, I think we can maybe get a similar effect with `TESTOPTS=--verbose bundle exec rake`?

javierjulio:
@deivid-rodriguez thanks! ❤️ No rush on this. It just surprised me how many warnings came up when I first ran the test suite. 😳 I felt this was something I could do. We definitely need to do it with Rails 6 on the horizon too.

Regarding the addition of `minitest-reporters`, that was only temporary to help me because I was familiar with it. I have the commit marked as TEMP so my plan is to drop them before review/merge.

I did try the `TESTOPTS=--verbose bundle exec rake` option and various forms of that but it just outputs times, no test case/class names at all. 😞 So for example, I see a long list of just the following:

```
0.00 s = .
0.00 s = .
0.01 s = .
0.01 s = .
0.00 s = .
0.00 s = .
0.00 s = .
0.00 s = .
0.00 s = .
```

deivid-rodriguez:
Yes, I just run the test suite and it's overwhelming, wow. It's so much better after this PR.

Regarding `minitest-reporters`... sorry, I didn't see the TEMP mark, looks good then! In any case, I think the weird output you're seeing is because of some incompatibility between the gems, if you fully remove the `minitest-reporters` gem, `TESTOPTS=--verbose bundle exec rake` should give you a better output like

```
(...)
MountableEngineTest#test_route_prefix_do_not_include_engine_name = 0.00 s = .
MountableEngineTest#test_route_prefix_present_when_parent_module_is_not_a_engine = 0.00 s = .
DestroyActionBaseTest#test_show_flash_message_when_user_can_be_deleted = 0.00 s = .
DestroyActionBaseTest#test_show_flash_message_with_javascript_request_when_user_cannot_be_deleted = 0.01 s = .
DestroyActionBaseTest#test_redirects_to_the_resource_if_cannot_be_destroyed = 0.01 s = .
DestroyActionBaseTest#test_show_flash_message_with_javascript_request_when_user_can_be_deleted = 0.00 s = .
DestroyActionBaseTest#test_the_requested_user_is_destroyed = 0.00 s = .
DestroyActionBaseTest#test_redirects_to_users_list = 0.00 s = .
DestroyActionBaseTest#test_show_flash_message_when_user_cannot_be_deleted = 0.00 s = .
StrongParametersTest#test_with_permitted_and_resource_methods = 0.00 s = .
StrongParametersTest#test_permitted_params_from_create = 0.00 s = .
StrongParametersTest#test_permitted_params_from_new = 0.00 s = .
StrongParametersTest#test_permitted_params_from_update = 0.00 s = .
StrongParametersIntegrationTest#test_permitted_params_from_new = 0.00 s = .
StrongParametersIntegrationTest#test_permitted_empty_params_from_new = 0.00 s = .
StrongParametersIntegrationTest#test_permitted_params_from_create = 0.00 s = .
StrongParametersIntegrationTest#test_permitted_params_from_update = 0.00 s = .
PolymorphicCompanyTest#test_polymorphic_helpers = 0.00 s = .
PolymorphicCompanyTest#test_the_requested_employee_is_destroyed_on_destroy = 0.00 s = .
PolymorphicCompanyTest#test_update_the_requested_object_on_update = 0.00 s = .
PolymorphicCompanyTest#test_expose_all_employees_as_instance_variable_on_index = 0.00 s = .
PolymorphicCompanyTest#test_expose_a_new_employee_on_new = 0.00 s = .
PolymorphicCompanyTest#test_expose_a_newly_create_employee_on_create = 0.00 s = .
PolymorphicCompanyTest#test_expose_the_requested_employee_on_show = 0.00 s = .
PolymorphicCompanyTest#test_expose_the_requested_employee_on_edit = 0.00 s = .
EngineLoadErrorTest#test_does_not_crash_on_engine_load_error = 0.00 s = .
```

javierjulio:
@deivid-rodriguez you're right! Great catch! I had just commented out the require/use calls in the test_helper but I didn't actually remove the gem. I didn't think it would still apply but it does. I'll drop that commit now and just use the verbose flag. Thanks! 👍🏻

javierjulio:
@deivid-rodriguez I've tried undefining methods to resolve the parent/parent? warnings but I'm not having any luck, sorry. 😞 Any chance you can help out?

deivid-rodriguez:
Sure! Let me have a look!

javierjulio:
@deivid-rodriguez that did it! Amazing. 😍 I'm glad I asked because I definitely wouldn't have noticed that at all! Such a simple solution too. 👌🏻 Thanks a lot!

javierjulio:
@deivid-rodriguez tomorrow I'll start squashing related commits and then look into moving the global test resources/routes into specific tests. Thanks again! 👍🏻

deivid-rodriguez:
I'd like to test this against activeadmin since I don't know this library well enough to be sure it won't break anything... It seems to make sense, but still...

javierjulio:
Ok great, that sounds good. 👍🏻

test/test_helper.rb
```diff
@@ -30,10 +30,82 @@ class ApplicationController < ActionController::Base; end
 
 InheritedResources::Routes = ActionDispatch::Routing::RouteSet.new
 InheritedResources::Routes.draw do
-  get ':controller(/:action(/:id))'
-  get ':controller(/:action)'
-  resources 'posts'
-  root :to => 'posts#index'
+  # aliases_test.rb
+  resources :students
```
deivid-rodriguez: I see why we need to do this to avoid all the warnings, and it was such a great job to track which route is needed by each test. :+1: :+1:

It'd be great if each test using each route could define & undefine the route on the fly. It would make the tests easier to follow, without having the make all these notes about which test uses which route (comments that definitely make sense in the current situation). But if we can't think of something now, we can leave that improvement for another time.

deivid-rodriguez: Hah, so you planned to [do this](https://github.com/activeadmin/inherited_resources/pull/459#issuecomment-434509742) already? :smile: :+1:

javierjulio: Thanks! 👍🏻 Some of the tests can be hard to follow and have some crazy resource setups. 😳 I thought I'd never figure them out as I'm not too familiar with singular resources but eventually I got it all to work. 😅 After I had worked my way through several test files I realized that as I added resources for each test file and ran the test suite, I had previous resources from other tests enabled. So I went back to each test file, I commented out all resources and made sure that each test file I listed there, only had what resources that file actually depended on, even if it meant it was duplicated here. This way it would be easier to start moving the resource definitions into each test file and not have them be global.

Haha yes I had planned to do it for awhile I'm just not sure yet how I will. At most I was going to try to make building routes into a simple helper method that can be called on setup and then another to remove/reset all routes on teardown.

As a side note, I tried experimenting by changing the tests to be integration tests but ran into what I think is/was a dependency issue. We might not have the right dependencies to enable that. Not that the tests have to change but it came to mind as we rely on rails-controller-testing.

deivid-rodriguez: Nice strategy to approach this, and also good idea on how to move each route to its own test! :+1:

javierjulio:
@deivid-rodriguez this is ready for review. I inlined in each test only the routes it requires. Would love to know what you think! I edited the commits so we now just have one that removes the global test routes.

Note that [the builds for Rails 5.0 emit a deprecation warning](https://circleci.com/gh/activeadmin/inherited_resources/549) that I don't see on other versions. I'm not sure if this warning is important for us to fix.

> DEPRECATION WARNING: Comparing equality between `ActionController::Parameters` and a `Hash` is deprecated and will be removed in Rails 5.1. Please only do comparisons between instances of `ActionController::Parameters`. If you need to compare to a hash, first convert it using `ActionController::Parameters#new`. (called from build_resource at /home/circleci/project/lib/inherited_resources/base_helpers.rb:52)

Despite this instance saying line 52, other warnings show different line numbers. Some state a different file too, for example, some report:

> (called from method_missing at /home/circleci/project/test/base_test.rb:66)
> (called from update_resource at /home/circleci/project/test/customized_base_test.rb:29)

Should we address this?

test/test_helper.rb
javierjulio: Resolving this as I dropped all those commits now. I have a commit where I remove all global test routes as they have been inlined in each test case. Thanks!

lib/inherited_resources/polymorphic_helpers.rb
```diff
@@ -120,8 +120,11 @@ def parent_class
       #
       def parent
         if parent_type
-          p = instance_variable_get("@#{parent_type}")
-          p || instance_variable_set("@#{parent_type}", association_chain[-1])
+          if instance_variable_defined?("@#{parent_type}")
+            instance_variable_get("@#{parent_type}")
+          else
+            instance_variable_set("@#{parent_type}", association_chain[-1])
+          end
```
deivid-rodriguez: Are these really equivalent? Previously, if the instance varaible was `nil`, we were setting it, now we're just returning `nil`, right?

Maybe something like the following would be safer to preserve the behavior?

```ruby
p = instance_variable_defined?("@#{parent_type}") && instance_variable_get("@#{parent_type}")
p || instance_variable_set("@#{parent_type}", association_chain[-1])
```

deivid-rodriguez (commented):
Nice cleanup. I only left one comment!

deivid-rodriguez:
@javierjulio Nice job. I wouldn't care about warnings on older versions.

javierjulio:
@deivid-rodriguez ok great thanks! 👍🏻 I'll get this updated for you shortly!

lib/inherited_resources/polymorphic_helpers.rb
javierjulio: I think I see what you mean, because the submitted change performed an existence check which it could exist but with a value of `nil` so it wouldn't be the same behavior. Sure, I have no issue with changing it to that. 👍🏻 Safe is best. Thanks for taking the time! Let me change that now.

deivid-rodriguez: Exactly! It might not have any real effect since the test suite didn't catch it but as you said, safe is best! :+1:

javierjulio: Yes I agree! I made the change so this is ready.

javierjulio:
@deivid-rodriguez I rebased again after merging in #469. Thanks! 👍🏻

deivid-rodriguez (approved):

deivid-rodriguez:
Thanks so much @javierjulio!

javierjulio:
@deivid-rodriguez thanks to you! You helped me out with some of the harder issues I had here. Thanks again for all the help and support! 👏🏻❤️

javierjulio:
@deivid-rodriguez regarding commit 00d1c0bb7054b6994ca4fd7b7ac6bdd4e5c4a9d1 which ran the cases in `belongs_to_test.rb` in order have to do with the test cases that deal with redirects. I compared that test case against another similar one and I just wasn't noticing any difference that would suggest why it would be failing randomly.

I noticed the redirect test cases and once I comment those out, several test runs (all random) are all green. Looking more carefully I think it makes sense since it modifies the controller to exclude certain actions. That's why we get failures that there is no index or show action on random test runs.

PR #460, opened 2018-10-27
General update to gem template files

javierjulio:
Since this library has been around for so long, the files that are generated when creating a gem (gemspec, Rakefile, etc.) have since changed. I've updated these files to be inline with what the same files would be from a newly generated gem. This has some nice advantages where in tests we can just use `require 'test_helper'` instead, for example. 👍🏻

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,12 +1,12 @@
-# encoding: UTF-8
-
 require 'bundler/gem_tasks'
 require 'rake/testtask'
 require 'rdoc/task'
 
 desc 'Run tests for InheritedResources.'
 Rake::TestTask.new(:test) do |t|
-  t.pattern = 'test/**/*_test.rb'
+  t.pattern = "test/**/*_test.rb"
+  t.libs << "test"
+  t.libs << "lib"
   t.verbose = true
 end
 
@@ -21,4 +21,4 @@ Rake::RDocTask.new(:rdoc) do |rdoc|
 end
 
 # Make test the default task.
-task :default => :test
+task default: :test
--- a/inherited_resources.gemspec
+++ b/inherited_resources.gemspec
@@ -1,5 +1,5 @@
-# -*- encoding: utf-8 -*-
-$:.push File.expand_path("../lib", __FILE__)
+lib = File.expand_path("lib", __dir__)
+$LOAD_PATH.unshift(lib) unless $LOAD_PATH.include?(lib)
 require "inherited_resources/version"
 
 Gem::Specification.new do |s|
--- a/inherited_resources.gemspec
+++ b/inherited_resources.gemspec
@@ -7,7 +7,7 @@ Gem::Specification.new do |s|
   s.version     = InheritedResources::VERSION.dup
   s.platform    = Gem::Platform::RUBY
   s.summary     = "Inherited Resources speeds up development by making your controllers inherit all restful actions so you just have to focus on what is important."
-  s.homepage    = "http://github.com/activeadmin/inherited_resources"
+  s.homepage    = "https://github.com/activeadmin/inherited_resources"
   s.description = "Inherited Resources speeds up development by making your controllers inherit all restful actions so you just have to focus on what is important."
   s.authors     = ['José Valim', 'Rafael Mendonça França']
   s.license     = "MIT"
--- a/test/aliases_test.rb
+++ b/test/aliases_test.rb
@@ -1,4 +1,4 @@
-require File.expand_path('test_helper', File.dirname(__FILE__))
+require 'test_helper'
 
 def plain_text
   ActionPack::VERSION::MAJOR >= 5 ? :plain : :text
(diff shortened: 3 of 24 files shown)


javierjulio:
@deivid-rodriguez no rush on this. I'm not sure what else I can do here so I'm submitting for review. If you have any ideas or suggestions let me know. Thanks!

inherited_resources.gemspec
```diff
@@ -1,5 +1,6 @@
-# -*- encoding: utf-8 -*-
-$:.push File.expand_path("../lib", __FILE__)
+# coding: utf-8
+lib = File.expand_path('../lib', __FILE__)
```
deivid-rodriguez: The magic comment shouldn't be necessary these days... Also, [master's bundler](https://github.com/bundler/bundler/blob/fbb9c2dada3d98ce7932f9767ea3b2bdbba04eef/lib/bundler/templates/newgem/newgem.gemspec.tt#L6) uses `__dir__` here, which is a little simpler :)

inherited_resources.gemspec
```diff
@@ -14,7 +15,9 @@ Gem::Specification.new do |s|
 
   s.rubyforge_project = "inherited_resources"
 
-  s.files         = Dir["app/**/*", "lib/**/*", "README.md", "MIT-LICENSE"]
+  s.files         = `git ls-files -z`.split("\x0").reject { |f| f.match(%r{^(test|spec|features)/}) }
```
deivid-rodriguez: I'm aware bundler's default is to shell out to `git`. My preferred solution though is the current one, since I find it more explicit and it doesn't cause errror messages when building the gem on systems without `git` installed (like very light docker images). Bundler's default approach is meant for non experienced users creating gems for the first time, because it prevents them from accidentally shipping generated (not source controlled) files with the gem.

inherited_resources.gemspec
```diff
@@ -14,7 +15,9 @@ Gem::Specification.new do |s|
...
+  s.bindir        = "exe"
+  s.executables   = s.files.grep(%r{^exe/}) { |f| File.basename(f) }
```
deivid-rodriguez: Even if these are bundler's default, I don't think we need the prior two lines because this gem has no CLI.

Rakefile
```diff
@@ -1,12 +1,12 @@
...
 require 'bundler/gem_tasks'
 require 'rake/testtask'
 require 'rdoc/task'
 
 desc 'Run tests for InheritedResources.'
 Rake::TestTask.new(:test) do |t|
-  t.pattern = 'test/**/*_test.rb'
+  t.libs << "test"
+  t.libs << "lib"
+  t.test_files = FileList['test/**/*_test.rb']
```
deivid-rodriguez: I'm unsure of the difference between `test_files` and `pattern`, but if this is what the default gem skeleton uses, I'm good with it! Adding `test` and `lib` to the LOAD_PATH seems definitely a good idea :+1:.

deivid-rodriguez (commented):
@javierjulio Thanks for doing this! I made a few comments for things I don't necessarily agree with but I like the spirit of the PR. :+1:

javierjulio:
@deivid-rodriguez no problem at all! I'm more than happy to make any changes. I'll tackle those for you. Thanks again!

Rakefile
javierjulio: Yes neither am I. I just went off of past experience making a gem where I noticed regardless of the file location, I could just simply have `require 'test_helper'` which I liked. I figured this had to do with it which it seems too. Thanks!

inherited_resources.gemspec
javierjulio: Sounds good to me! I will remove them. 👍🏻

inherited_resources.gemspec
javierjulio: That's what I figured. Great, I will update it to follow that! Thanks!

inherited_resources.gemspec
javierjulio: I have no issue with that too. I will leave it as it was then. Thanks!

javierjulio:
@deivid-rodriguez please let me know if I missed anything but this should now have all the changes requested. Thanks! ❤️

deivid-rodriguez (approved):

Rakefile
deivid-rodriguez: Actually, I think I have a reason to prefer `pattern` over `test_files`. RSpec also uses a [`pattern` option](https://github.com/rspec/rspec-core/blob/9d1d0d6aeb8dc87ff4e584cdcf9e8b1a234da2ae/lib/rspec/core/rake_task.rb#L35) for the exact same thing, so by using it here too, a new developer doesn't need to learn a new API, no matter what test framework is in use. If you agree, we can change it back and instead propose to change the template in bundler :)

javierjulio: @deivid-rodriguez sure that sounds good to me! 😄 I went ahead and made the change. I'll merge this first once the build passes. Thanks! 👏🏻

deivid-rodriguez:
Wowowooooow :tada:

PR #464, opened 2018-10-27
Enable rubocop

deivid-rodriguez:
While reviewing #460 I saw several whitespace changes being introduced with the PR. So I'm enabling a rubocop rule to ensure that we don't reintroduce trailing whitespaces never again in the future!


javierjulio:
@deivid-rodriguez sorry, since I merged in other changes there is a conflict here now. 😞 No rush on this but let me know when you get to it. Happy to get this in! 👍🏻

javierjulio:
@deivid-rodriguez since I'm updating the tests, I'd like to address some minor inconsistencies, for example, instead of defining a class or module in one line using semi colons (`class User; end`) to instead require it be multi line. Is this something we could target with Rubocop? Would you be open to that? I don't think you should make that change here though since I'll be updating the tests, but if we can do that I can enable it when I do that update.

deivid-rodriguez:
@javierjulio I rebased this PR and added a few more cops for style changes that I saw in your other PRs. Only the last one `Layout/EmptyLinesAroundClassBody` does not match your changes. This is because I tried to apply the style you were introducing (empty line before class ending), but it lead to a lot of offenses, so I preferred to keep the style that's currently there (which is rubocop's default anyways).

Regarding what you mentioned about dummy class definitions, I like multiline over single line with semicolon as well, so I guess we could add that too, although it's not the most common style from what I've seen so there might be many offenses.

deivid-rodriguez:
I checked rubocop support for empty class definitions, and it's not there yet, although Gitlab seems to have implemented something. We would need an option to enforce the "bad" style :smiley:. Have a look at https://github.com/rubocop-hq/rubocop/issues/4082.

deivid-rodriguez:
> I don't think you should make that change here though since I'll be updating the tests, but if we can do that I can enable it when I do that update.

I'm not sure I understood the reason for not doing it here, is it merge conflicts for updating the same region of code? I think that'd be a little more work to rebase but it will free the other PR from unrelated stylistic-only changes, so it's good.

This is ready, by the way!

javierjulio:
@deivid-rodriguez yes I think it was regarding merge conflicts. I don't really remember now. 😂 I believe that I expected to update a lot of those files but I think its fine. I will be splitting up at least one set of changes from #459 anyway. Also I probably thought you wouldn't get to this sooner. Either way its no problem. Thanks! 👍🏻

javierjulio (approved):
Thanks so much for doing this! 👍🏻 It's really helpful to have. I'll merge this in as is. Do you think you can set up IR to run Rubocop as a lint style build like you did for ActiveAdmin? Or perhaps something similar here?

deivid-rodriguez:
Absolutely, I meant to do it in this PR, if I didn't it's because I forgot :). I'll do a follow up PR!

javierjulio:
Sounds good thanks! ❤️

PR #465, opened 2018-10-28
Add CircleCI Setup

javierjulio:
This is a quick setup of the same test suite we run with Travis but now with CircleCI to compare the result. I've shamelessly taken @deivid-rodriguez's work (thanks!) from ActiveAdmin and added it here to see how build times improved.


javierjulio:
Wow. Each build ran in **less than a minute each on CircleCI**, yes even JRuby builds. Recent Travis builds suggest that they regularly take a total of 11-13 minutes. This same build here in CircleCI **took less than 3 minutes** to complete! The second build then [took less 2 minutes](https://circleci.com/workflow-run/87550d85-8dbd-48c3-aa98-83c36f6c4405). So with CircleCI we'd be having an overall build time of 2-3 minutes.

javierjulio:
@deivid-rodriguez the improvement is quite noteworthy that I think this is worth doing. Let me know what you think. If you're good with it I can make a commit to remove Travis builds. For now I've left it so you can also see a comparison here.

Also after enabling CircleCI on this repository it seems to have triggered this for the master branch rather than just keeping it only for this PR for now. Sorry about that. That was unintentional.

.circleci/config.yml
```diff
@@ -0,0 +1,230 @@
...
+  - *copy_current_gemfile
+  - *restore_cache
+  - *install_dependencies
+  - *save_cache
+  - *run_tests
+
+jobs:
+  ruby23rails50:
+    docker:
+      - image: circleci/ruby:2.3.7-node-browsers
```
deivid-rodriguez: I don't think we need the `-browser` variants here because there are no browser specs in this gem, and I'm not fully sure we need the `-node` variants either. Can you check?

deivid-rodriguez (commented):
Looks good to me @javierjulio, only added one small comment, to see if we can use lighter base images.

.circleci/config.yml
javierjulio: Sure! I'll give it a try without both. I couldn't confirm if one exists without `-browsers` and `-node` but I think it does.

javierjulio: It worked! That's awesome, thanks! Fastest build yet. I'll squash that change into the first commit.

javierjulio:
@deivid-rodriguez we definitely can use the lighter base images! I made the switch and it works great. Super fast. Let me know if you are good with this and I'll get it in, thanks! ❤️

deivid-rodriguez (approved):

javierjulio:
Thanks David! 👍🏻

PR #466, opened 2018-10-30
More rubocop

deivid-rodriguez:
This is a follow up to #464.

It comes from having a look at the diff in #459, and globally fixing the style only changes included in there. Also, I added rubocop as part of the default task so it's run in CI.

Downsides:
* This PR is big.
* This PR will create a lot of conflicts with #459.

Upsides:
* Consistency formatting all around the library.

What do you think @javierjulio? I can help with rebasing #459 if we do this.


lib/inherited_resources/base_helpers.rb
```diff
@@ -132,12 +132,12 @@ def association_chain
...
-      #
-      # def flash_interpolation_options
-      #    { }
-      # end
+    # Overwrite this method to provide other interpolation options when
+    # the flash message is going to be set.
+    #
+    # def flash_interpolation_options
+    #    { }
+    # end
```
deivid-rodriguez: The indentation here was correct and it seems like a bug in rubocop that it's being changed. I'll revert these changes and report upstream.

javierjulio: Sounds good thanks!

javierjulio (commented):
I love it! 👏🏻 There were some small changes I wanted to get in from #459 separately and they will in tie with this. I think it will help us avoid noisier merge conflicts and keep smaller diffs. 👍🏻 

From #459 I will cherry pick the commits that do the following:

* Add `request_params(id: '42')` for single resource calls that require it and update method definition to support additional options. I'll update existing commits with the new syntax.
* Correct indentation for test_redirect_to_index_url_after_update
* Remove duplicate test case test_url_helpers_on_nested_polymorphic_belongs_to

What do you think?

lib/inherited_resources/base_helpers.rb
javierjulio: Sounds good thanks!

deivid-rodriguez:
I think all the changes in #459 that either remove duplicate tests, or rename duplicate tests so that they are actually run can be proposed separately independently from this, yeah :+1:. Regarding the commit fixing indendation, I think it's better to fix that in here, globally. And regarding the calls to `request_params`, I'm not sure.

Either way, this PR should be easy to recreate on top of any master, since its commits are mostly automatically generate via `bundle exec rubocop --auto-correct`, so feel free to propose anything you think it's ready separately, and let's land that first.

javierjulio:
@deivid-rodriguez ok sounds good! Thanks. 👍🏻 I just submitted #467 with what I thought would help. I figured you were doing auto correct so I think that will make it easier for both of us. After #467 and this change, the only big merge conflict I should have to deal with is the commit 33de5724c29e9a8b79f134372a034802889ba58f.

deivid-rodriguez:
I think it also makes sense to include 33de572 in #467!

javierjulio:
@deivid-rodriguez ah yes that makes sense! I thought I'd deal with it but since its a bigger diff and you autocorrect, it would be better to get it in sooner. I'll do that. Thanks David! 🙏🏻

javierjulio:
@deivid-rodriguez let me know once this is rebased and ready to go, thanks!

lib/inherited_resources/base_helpers.rb
deivid-rodriguez: Rebased but still pending to remove these changes, I'll ping you when it's done!

deivid-rodriguez: @javierjulio Done!

javierjulio: Ok great thanks! 👍🏻 I'll take a look now.

javierjulio (approved):
Looks great! Thanks again for setting this all up. 👌🏻❤️ I'll merge in and rebase in #459 which should be easier now.

PR #467, opened 2018-10-30
General updates to test cases

javierjulio:
This is a collection of changes that started in #459 but to keep that diff small and to ease in the changes from #466, I think it would help to get this in first. I made sure to undo style based changes I did so #466 should have in general less conflicts once this goes and #466 is rebased.


javierjulio:
Hmm seems the service is down where prebuilt docker images are downloaded from. I'm seeing all 503 Service Unavailable errors. I'll rerun the workflow later today.

deivid-rodriguez:
Yeah, looks like the docker registry is having issues, let's retry later :+1:

deivid-rodriguez:
Incident statuses at https://status.circleci.com/ and https://status.docker.com.

javierjulio:
@deivid-rodriguez I reran this a few times and its all good to go now! Docker Hub Registry is back up and the builds are passing. Thanks!

test/test_helper.rb
```diff
@@ -17,8 +17,9 @@
 require 'rails-controller-testing'
 Rails::Controller::Testing.install
 
-def request_params(params)
-  { params: params }
+# FIXME: update tests now to inline this
+def request_params(params, **options)
+  { params: params }.merge(options)
```
deivid-rodriguez: Is this still needed now that we dropped Rails 4 support?

test/customized_redirect_to_test.rb
```diff
@@ -18,12 +18,12 @@ def test_redirect_index_url_after_create
...
-     assert !PostsController.respond_to?(:show)
-     put :update
-     assert_redirected_to 'http://test.host/posts'
-   end
+  def test_redirect_to_index_url_after_update
+    Post.stubs(:find).returns(mock_machine(:update_attributes => true))
+    assert !PostsController.respond_to?(:show)
+    put :update
+    assert_redirected_to 'http://test.host/posts'
+  end
```
deivid-rodriguez, line 26: I would drop the commit fixing the indentation, I can take care of it in the rubocop PR. But if you'd like to keep it it's ok too! :+1:

test/test_helper.rb
javierjulio: No it's just temporary. I still need to inline this in all tests which I will I just haven't gotten around to it and mostly because of #466 which touches a lot of files. I can do that now though, it's no problem. 

What do you think is best? Should I do that update now here or merge this in as is, then we merge #466 and I submit a new PR to replace `request_params` in all our tests and inline the params locally?

test/customized_redirect_to_test.rb
javierjulio: Yes I was unsure too. I figured it was best to fix indentation just for this test case (I undid the indentation fix for another piece in this file) as I had to update this test anyway. If you feel strongly about not having this indentation fix here though I will remove.

test/test_helper.rb
deivid-rodriguez: I see. I mentioned Rails 4 because it was mentioned in the commit message and it seemed no longer relevant.
 
> What do you think is best? Should I do that update now here or merge this in as is, then we merge #466 and I submit a new PR to replace request_params in all our tests and inline the params locally?

I wasn't sure either, so I ended up inlining it myself in https://github.com/activeadmin/inherited_resources/pull/467/commits/8c50ee714ecb91e098a2ebf8251d35355aae435a :smiley:

test/customized_redirect_to_test.rb
deivid-rodriguez: No, I don't feel strongly, we can update it here! :+1:

test/test_helper.rb
deivid-rodriguez: I think its good now, only thing is that the diff has a mix of hash key styles that I find a bit distracting. Can be fixed, but no big deal.

deivid-rodriguez (approved):

test/test_helper.rb
javierjulio: Ok great thanks! 👍🏻 I'll get this in now.

javierjulio:
Thanks David! 👍🏻

PR #468, opened 2018-11-06
Bump loofah version in lock files

deivid-rodriguez:
Not a real vulnerability under our responsability (users should upgrade the dependency on their final apps), but let's bump it anyways to "fix" the security alerts.


javierjulio (approved):
Yes agreed, we should update. Thanks! 👍🏻

PR #470, opened 2018-11-14
Update test to run in random order

javierjulio:
While working on #459 I noticed that the `belongs_to_test.rb` had specified the tests to run in order rather than randomly which suggests there is global state affecting other test cases. The issue is that in the redirect related test cases, the controller is modified to include explicit actions where the other tests rely on the default. 

I couldn't figure out how to undo that change, so I split out the redirect cases into a separate test with its own named controller. I'm open to suggestions though if this can be resolved by keeping the test cases where they were or any other changes or improvements.

### DIFF
--- a/test/belongs_to_test.rb
+++ b/test/belongs_to_test.rb
@@ -118,8 +118,6 @@ class BelongsToTest < ActionController::TestCase
   end
 
   def test_helpers
-    mock_post.stubs(:class).returns(Post)
-
     Comment.expects(:scoped).returns([mock_comment])
     get :index, params: { post_id: '37' }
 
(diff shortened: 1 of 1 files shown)


javierjulio:
@deivid-rodriguez let me know what you think about this solution. You might have a better idea. I'm happy to make any improvements or changes. This is the best I could come up with.

Also I ran this against several failed seeds and did full test suite runs (all random order) and there are no failures now with this test running in the default (random) order.

deivid-rodriguez (approved):
Nice!

javierjulio:
Awesome thanks! 👍🏻 I'll get this in now.

PR #471, opened 2018-11-15
Bump rack

deivid-rodriguez:
Because it does not hurt, and shushes github security alerts.


javierjulio (approved):
Nice! Like that rake task. I didn't notice the security alert until now. Thanks! 👍🏻

Issue #472, opened 2018-12-10
Unable to autoload constant Base


sergiogomez:
This issue was originally reported at https://github.com/activeadmin/activeadmin/issues/2908

Current workarounds (from the engine) are:

- Rename `Base` classes
- Add an empty `Base` class

```
class Base
end

module MyCustomEngine
  class Engine < ::Rails::Engine
```

I guess it could be fixed rescuing `NameError, LoadError` at method `initialize_resources_class_accessors!` from `inherited_resources/lib/inherited_resources/class_methods.rb` (as done at https://github.com/activeadmin/inherited_resources/pull/416)

brianjlandau:
I ran into this trying to get `activeadmin` and `activeadmin-mongoid` to work. I put this in my `config/application.rb`:

```ruby
module AppName
  class Application < Rails::Application
    initializer :load_inherited_resources_base, before: 'active_admin.mongoid.resource_controller' do
      require 'application_helper'
      require 'devise_helper'
      require 'preview_helper'
      require 'application_controller'
      ir_gem_path = Bundler.rubygems.find_name('inherited_resources').first.full_gem_path
      require "#{ir_gem_path}/app/controllers/inherited_resources/base"
    end
  end
end
```

Hopefully anyone looking for a solution to this finds this :partying_face:

maximevaillancourt:
I'm running into this, though very intermittently / randomly, and I'm struggling to reproduce the exception.

It seems like there are two associated exceptions in my case (one shadows the other):
- `uninitialized constant InheritedResources::Actions`
- `uninitialized constant InheritedResources::Base`

In case others see this too, here are two very similar issues that discuss this kind of problem (which seems related to Ruby's `autoload`):
- https://github.com/rails/rails/issues/45600
- https://github.com/formtastic/formtastic/issues/1364

istana:
I was able to fix this error by using `Rails.application.config.to_prepare` in Rails with Zeitwerk:

~~~
initializer :active_admin_setup do
  Rails.application.config.to_prepare do
    ....
  end
end
~~~

PR #473, opened 2018-12-10
Rescue LoadError in addition to NameError

sergiogomez:
This PR could fix this issue: https://github.com/activeadmin/inherited_resources/issues/472


rafaelfranca:
I think before fixing this by rescue exceptions we need to understand why this fails. Creating a class called `Base` should not affect the way a gem is loaded.

sergiogomez:
I agree, and it's related to ActiveAdmin, because it first fails initializing `ActiveAdmin::BaseController`.

PR #476, opened 2019-01-06
Update rubocop to 0.62.0

deivid-rodriguez:
It's the first version supporting MRI 2.6.0 so we'll need it to run CI against that ruby version.


javierjulio (approved):
Thanks! 👍🏻

PR #479, opened 2019-03-01
Add support for Rails 6 beta

ionut998:
Allow newer versions of actionpack and railties


javierjulio:
@ionut998 thanks!

@deivid-rodriguez I think it's a good time to set up a Rails 6 beta CI build, what do you think?

ionut998:
@javierjulio I just added the setup for Rails 6 beta CI build

test/gemfiles/Gemfile-Rails-6-0-beta
```diff
@@ -0,0 +1,10 @@
+source 'https://rubygems.org'
+
+gemspec :path => '../..'
+
+gem 'rails', '6.0.0.beta1'
```
javierjulio: Actually beta2 was just released a few days ago!

```suggestion
gem 'rails', '6.0.0.beta2'
```

javierjulio (commented):
Thanks for contributing the CI build as well. 👍🏻

javierjulio (approved):
Awesome, thanks a lot @ionut998! This looks great. 👍🏻

.circleci/config.yml
```diff
@@ -133,6 +133,16 @@ jobs:
...
+  ruby25rails60beta:
+    docker:
+    - image: circleci/ruby:2.5.1
+
+    environment:
+      BUNDLE_GEMFILE: test/gemfiles/Gemfile-Rails-6-0-beta
+
+    steps:
+      *test_steps
+
```
deivid-rodriguez, line 145: I guess we can add the whole matrix and test against all supported rubies since we are at it?

matheusvetor: We need also add ruby 2.6

matheusvetor (changes requested):

.circleci/config.yml
ionut998: @deivid-rodriguez according to the guideline Rails 6 requires ruby 2.5 or newer
https://edgeguides.rubyonrails.org/upgrading_ruby_on_rails.html#ruby-versions

deivid-rodriguez: You're absolutely right, my bad! This PR is going in then!

deivid-rodriguez:
Thanks @ionut998!

javierjulio:
Sorry I couldn't follow up @ionut998 but thanks a lot for this and the recent updates! Much appreciated. ❤️

ionut998:
Thanks @javierjulio, Happy to help
Hope to see this in a new release

inherited_resources.gemspec
```diff
@@ -20,7 +20,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 2.3'
 
   s.add_dependency("responders")
-  s.add_dependency("actionpack", ">= 5.0", "< 5.3")
-  s.add_dependency("railties", ">= 5.0", "< 5.3")
+  s.add_dependency("actionpack", ">= 5.0", "< 6.0")
+  s.add_dependency("railties", ">= 5.0", "< 6.0")
```
deivid-rodriguez, line 24: I was thinking about this a bit more, and I think this is a good strategy.

We allow any prereleases for now, but not the final version. We'll watch for compatibility from now on but hold off a new release until Rails 6.0 is released and we confirm we are compatible. Then bump the upper limit here to 6.1 for example and release a new version

PR #483, opened 2019-03-08
Jruby improvements in CI

deivid-rodriguez:
Little tweaks to jruby in CI.

PR #482, opened 2019-03-08
added missing test rails6beta against jruby92

matheusvetor:

PR #484, opened 2019-03-09
Reconcile jruby 9.2 + Rails 6.0 base docker image

deivid-rodriguez:
Follow up to #483 and #482. Use the same base image and tag as the other jruby images.


javierjulio (approved):
Awesome thanks!

PR #486, opened 2019-03-24
Improve gemspec

deivid-rodriguez:
This PR makes sure our gemspec builds without warnings. It fixes the following warnings:

```
WARNING:  description and summary are identical
WARNING:  open-ended dependency on responders (>= 0) is not recommended, use a bounded requirement, such as '~> x.y'
```

And also removes the `rubyforge_project` attribute since it currently does nothing and it will be deprecated soon.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,6 +33,11 @@ version: 2.1
     name: Make sure gemfiles are up to date
     command: bundle exec ruby -Itest -e 'require "gemfiles/lint.rb"; Minitest.run'
 
+.check_gemspec: &check_gemspec
+  run:
+    name: Make sure gemspec builds without warnings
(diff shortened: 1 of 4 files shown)


javierjulio (approved):
Awesome, thanks! 👍🏻 I was updating the ENVied gem yesterday to keep it current and I just tried now the build command on it and noticed the same issues there. Love how you automate a check using tests so I'll be doing the same! I didn't know about the open-ended dependency warning.

deivid-rodriguez:
Thanks! Yeah, for gems you don't want to be too restrictive with your dependency ranges, but keeping them under a sane range instead of just allowing everything is good. And helps bundler!

javierjulio:
Yes agreed! In the case for ENVied it was due to listing bundler and rake as development dependencies with no range but I'm not sure if that's needed (I've removed them) since they both come with Ruby now. I've implemented the same test case you did here too to automate that. ❤️ Always learning a lot from your changes, thanks! 👏🏻

PR #487, opened 2019-03-24
Cut 1.10.0.rc1

deivid-rodriguez:
I'd like to cut a version supporting Rails 6, and also release some refactorings on master such as #455.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
 # Master (unreleased)
 
+# Version 1.10.0.rc1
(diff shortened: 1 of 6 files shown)


javierjulio (approved):
Awesome, thanks! Let's release it! 👍🏻

javierjulio:
I noticed we are missing a build for Rails 5.2 so I'll see about getting that in for us.

EDIT: Ah yes the default build is for Rails 5.2 so nevermind.

PR #496, opened 2019-07-30
Configure dependabot

deivid-rodriguez:


javierjulio (approved):
Looks great, thanks! I wasn't sure if you were all done but I can merge. Let me know.

timoschilling (approved):
❤️

deivid-rodriguez:
Thanks both, let's try this out!

deivid-rodriguez:
From the initial batch of PRs, I see that dependabot does not assign the same meaning to "development" dependency in the case of gems as I would've expected.

As I understand it, in the case of gems, "development dependency" is everything that appears in the `Gemfile` (plus dependencies in the gemspec using `add_development_dependency`) whereas "production (runtime) dependency" is everything that appears in the gemspec using `add_dependency`.

Instead, `dependabot` seems to use the same meaning as in applications, so nothing is automerged because "no group" in the Gemfile means "production" dependency.

We might be able to workaround this by wrapping every gem in the Gemfile with `group :development`, but I will still report it to dependabot.

timoschilling:
I think automerge works only if the PR is approved before the CI is done or if the repo has no approve requirement.
(that's the learning from some docs and issue readings and the result of our evaluation on some private repos)

deivid-rodriguez:
Interesting, I didn't consider review requirements. However, in the PRs that we currently have opened, the PR description explains that 

> Dependabot will not automatically merge this PR because it includes a minor update to a production dependency.

So if your perception is right, there's still room for improvement in the explanation message.

@dependabot Any thoughts about this? Should I open an issue about it?

PR #510, opened 2019-07-31
Wrap Gemfile gems with development group

deivid-rodriguez:
To teach dependabot that in the case of gems, everything in the Gemfile is a development dependency.

Hopefully adding this makes dependabot automerge PRs? If it doesn't, it's probably because of the compulsory review requirement as suggested by @timoschilling.


javierjulio (approved):
Great thanks! Let's give this a try. 👍🏻

PR #512, opened 2019-07-31
Drop ruby 2.3 support

deivid-rodriguez:
So we get in line with activeadmin, and reduce the size of the CI matrix. And because 2.3 already reached its end of life.


javierjulio (approved):
Awesome, thanks for doing this! I'll get this in first before the other updates. You may have to update to resolve conflicts on some of them.

PR #515, opened 2019-07-31
Fix CI false negatives

deivid-rodriguez:
Some steps in CircleCI are actually failing, but the command was not exiting with the proper status, so the failures were "hidden". This PR:

* Fixes that issue so the correct status is reported.
* Fixes the failure by syncronizing the lockfiles.


javierjulio (approved):
Awesome thanks!

PR #531, opened 2019-08-21
Let Dependabot always handle security updates

deivid-rodriguez:
Dependabot doesn't currently create PRs for security updates if they are indirect production (runtime) dependencies (not direct dependencies of the gem, but dependencies of a runtime dependency). Like nokogiri here.
    
Since Github still creates security alerts for those, I'd like dependabot to handle these automatically too.
    
My guess is that dependabot classifies dependency types like this:
    
    ▾ all
    ├── ▾ direct
    │   ├── production
    │   └── development
    └── indirect
    
And not like this as I thought:
    
    ▾ all
    ├── ▾ production
    │   ├── direct
    │   └── indirect
    └── ▾ development
        ├── direct
        └── indirect
    
So specifying `type: production` won't suffice for this type of dependencies. Let's try with `type: all`.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -10,7 +10,7 @@ update_configs:
 
     allowed_updates:
       - match:
-          dependency_type: production
+          dependency_type: all
           update_type: security
 
     automerged_updates:
@@ -25,7 +25,7 @@ update_configs:
 
     allowed_updates:
       - match:
-          dependency_type: production
+          dependency_type: all
           update_type: security
 
     automerged_updates:
@@ -40,7 +40,7 @@ update_configs:
 
     allowed_updates:
       - match:
-          dependency_type: production
+          dependency_type: all
           update_type: security
 
     automerged_updates:
@@ -55,7 +55,7 @@ update_configs:
 
     allowed_updates:
       - match:
-          dependency_type: production
+          dependency_type: all
           update_type: security
 
     automerged_updates:


timoschilling (approved):

deivid-rodriguez:
:crossed_fingers:

deivid-rodriguez:
Well, this was intended so that Github will automatically create a PR to upgrade our vulnerable nokogiri, but it didn't work.

Funny enough, I ended up discovering that:

* Dependabot has exactly this feature enabled by default, so we probably won't need the previous two PRs since they are redundant.
* We are already using a [safe version of nokorigi](https://github.com/activeadmin/inherited_resources/blob/47acbb1b0859b68639b85ebe95f99e67c34e8e1d/Gemfile.lock#L110-L112) :rofl:. So this is just a false positive from Github, I think? I reported it and I'm waiting for a reply.

javierjulio:
Haha no problem at all! It makes sense. I looked into this yesterday as I received a GitHub security alerts email and it reported nokogiri as the safe version as you said, but with a less than operator which confused me since its not accurate. We have the safe version installed, not the previous version before it.

In regards to the first item, what additional changes do we need to make?

deivid-rodriguez:
Nothing. Maybe undoing the previous two configuration changes, since dependabot updates security updates in indirect dependencies by default. But it's not really necessary, I think?

javierjulio:
To undo would would mean removing the entire `allowed_updates:` block right? I haven't used Dependabot though on a gem, only app repos but with that the defaults have worked great. Whatever you prefer to do here (leave as is or undo) though I'm fine with. 👍🏻 Thanks again too for following up on this with what you found out!

deivid-rodriguez:
Yeah, I'll wait for a reply from Github regarding the false alerts, then decide what to do with the dependabot config (no big deal either way).

deivid-rodriguez:
Thanks @javierjulio!

PR #536, opened 2019-10-24
Drop Rails 5.0 and Rails 5.1 support

deivid-rodriguez:
Following activeadmin's lead.


javierjulio (approved):
Awesome thanks David! ❤️

PR #574, opened 2020-03-17
Update image versions in CI and add Ruby 2.7

javierjulio:
I've updated our CI config to use the latest image versions for all our builds. I've also added Ruby 2.7 to the build matrix with both a Rails 5.2 and 6.0 builds so we can see if any 2.7 specific warnings appear.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -71,7 +71,7 @@ version: 2.1
 jobs:
   lint:
     docker:
-      - image: circleci/ruby:2.6.3
+      - image: circleci/ruby:2.6.5
 
     environment:
       BUNDLE_GEMFILE: Gemfile
@@ -81,7 +81,7 @@ jobs:
 
   ruby24rails52:
     docker:
-      - image: circleci/ruby:2.4.6
+      - image: circleci/ruby:2.4.9
 
     environment:
       BUNDLE_GEMFILE: test/rails_52/Gemfile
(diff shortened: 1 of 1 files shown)


javierjulio:
I'll wait for CI to complete first and then update settings so we require the 2.7 builds.

.circleci/config.yml
```diff
@@ -131,7 +151,7 @@ jobs:
 
   jruby92rails52:
     docker:
-      - image: circleci/jruby:9.2.7.0
```
deivid-rodriguez, line 134: Latest jruby is 9.2.11.0, not sure if CircleCI has yet updated their images though.

deivid-rodriguez (approved):
Looks good!

.circleci/config.yml
javierjulio: Good to know! I checked using though rbenv so I missed that.

Looks like CircleCI does have 9.2.11 as I see it here: https://circleci.com/docs/2.0/docker-image-tags.json so good with updating to 9.2.11?

deivid-rodriguez: Yeah!

javierjulio: Ok great! I just pushed an update now. Thanks!

javierjulio:
Thanks David! 👍🏻 Everything passed so I'll merge and update the repo settings so we now require these builds.

deivid-rodriguez:
Thanks to you! :heart:

PR #638, opened 2021-01-30
Add rails_61 to Dependabot config

javierjulio:
This updates our Dependabot config to include our Rails 6.1 gemfile as I noticed that was missing but we should have it to stay in sync.

### DIFF
--- a/.dependabot/config.yml
+++ b/.dependabot/config.yml
@@ -14,6 +14,16 @@ update_configs:
           update_type: all
 
   - package_manager: ruby:bundler
+    directory: /test/rails_61
+    update_schedule: live
+    version_requirement_updates: increase_versions_if_necessary
+
+    automerged_updates:
+      - match:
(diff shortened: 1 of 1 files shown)


deivid-rodriguez:
@javierjulio Sorry I completely missed this. The  behavior regarding not updating rails seems fine to me. You're telling bundler to update nokogiri, so it keeps the locked version of rails since you haven't told bundler to update Rails.

Since I don't see any dependabot alerts now I assume this is fixed, but let me rebase this just in case.

javierjulio:
@deivid-rodriguez no problem! I checked each of the `Gemfile.lock` files and they all have the safe Nokogiri version installed so we are good. We should still get the Dependabot configuration change in though since it was missing the Rails 6.1 gemfile. I'll update the PR now for that.

deivid-rodriguez (approved):
Awesome, thanks!

PR #650, opened 2021-03-30
Drop ruby 2.4 support

deivid-rodriguez:


javierjulio (approved):
Thanks!

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,79 @@
...
+
+    strategy:
+      fail-fast: false
+
+    steps:
+      - uses: actions/checkout@v2
+
+      - uses: ruby/setup-ruby@v1
+        with:
+          ruby-version: 2.7.2
```
javierjulio, line 26: By default, Ruby 2.7.2 won't display deprecation warnings anymore. Do you know if we have it enabled in our Ruby 2.7 CI builds to display those? I've stuck with Ruby 2.7.1 in builds for my projects just to make sure to see deprecation warnings.

javierjulio (commented):
Thanks for updating and also doing the GitHub Actions migration! All around much better. I only have one concern with the Ruby 2.7 build.

.github/workflows/ci.yaml
deivid-rodriguez: I recall something around that and keyword arguments but not sure about the specifics. I guess we can run tests against 2.7.1 and fix any deprecation warningsShould I run tests with 2.7.1 and fix any warnings that come up.

deivid-rodriguez: I haven't enabled anything so I don't think we're currently displying them, no.

javierjulio: I'm not sure what's best but am flexible with using either version. If we do decide to go with Ruby 2.7.2 though, [we should enable deprecation warnings](https://www.ruby-lang.org/en/news/2020/10/02/ruby-2-7-2-released/). What do you prefer? We'd want to do this for our other repos too otherwise we may be hiding possible deprecation warnings.

deivid-rodriguez: I think we should enable them and stick with 2.7.2.

PR #668, opened 2021-05-06
Some development updates

deivid-rodriguez:
A dependabot PR updated our Rails 5.2 Gemfile to Rails 6.1: #663.

This PR restores that Gemfile to test the Rails version it's supposed to test, and tries to prevent the same upgrade in the future.

It also adds other minor updates.


javierjulio (approved):
Love it, thank you! 🙌🏻

PR #708, opened 2021-12-21
Test against latest Ruby versions

tagliala:
Also skip testing Ruby 2.7 against Rails 5.2 (ref: rails/rails#38426)

As requested at activeadmin/arbre#314


tagliala:
@Mergifyio refresh

javierjulio (approved):
Thanks!

PR #709, opened 2021-12-21
Drop support for Ruby 2.5

alejandroperea:
Ruby 2.5 arrived to EOL more than 8 months ago. I think we can remove it from the test matrix and drop the support.

### DIFF
--- a/.github/mergify.yml
+++ b/.github/mergify.yml
@@ -6,9 +6,6 @@ pull_request_rules:
     conditions:
       - author~=^dependabot(|-preview)\[bot\]$
       - check-success=lint
-      - check-success=test(ruby-2.5, rails_52)
-      - check-success=test(ruby-2.5, rails_60)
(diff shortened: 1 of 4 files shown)


deivid-rodriguez (approved):
👍

rafaelfranca:
I'm usually very agressive with dropping Ruby support but if we do here should not we do in all the other active admin repos?

deivid-rodriguez:
Yes, that would be the plan.

javierjulio (approved):
We already started to remove Ruby 2.5 support in other ActiveAdmin org gems, in fact @alejandroperea contributed that same change in https://github.com/activeadmin/activeadmin/pull/7236 (thanks!)

@alejandroperea do you mind rebasing and pushing again?

alejandroperea:
@javierjulio Rebased! 😄

javierjulio:
@alejandroperea awesome, thanks again!

PR #711, opened 2022-01-19
Remove upper bound dependency limits from gemspec

rafaelfranca:
All that was needed to support Rails 7 in this gem was to update the
gemspec. This means that even in the most likely release to break
anything, a major release, we didn't needed to worry about breaking this
gem.

This upper bound is just creating unnecessary work for the
maintainers of this gem, which would need to release it without any code
change, and the users, which need to open issues asking for new releases.

Let's allow our users to try any version of our dependencies and report
real issues if they find them, and save everyone's time just removing
the upper bound.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Master (unreleased)
 
+* Remove upper bound dependency limits from gemspec.
 * Allow using Rails 7.1.
 
 ## Version 1.13.1
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,10 +2,10 @@ PATH
   remote: .
   specs:
     inherited_resources (1.13.1)
-      actionpack (>= 6.0, < 7.2)
-      has_scope (~> 0.6)
-      railties (>= 6.0, < 7.2)
-      responders (>= 2, < 4)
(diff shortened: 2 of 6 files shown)


javierjulio (commented):
Thanks @rafaelfranca! I understand the reasoning and am ok with either removing or specifying upper bounds. Why did you remove the `gemspec_lint` test? That would help us confirm no warnings. Other than thought no concerns with the change. There are merge conflicts.

rafaelfranca:
> Why did you remove the `gemspec_lint` test? That would help us confirm no warnings

I removed because RubyGems show warnings for dependencies without upper bound. I wish I could just disable that warning so we could checking the others, but that is not the case

deivid-rodriguez:
I agree with your reasoning too @rafaelfranca. I would like to work on making RubyGems gemspec warnings configurable in the future. But for now it seems easy enough to ignore this particular warning from the output, instead of removing everything? I'm happy to work on it myself if you don't have the time.

rafaelfranca:
You mean in the test? I can totally do it. I'll update the PR to ignore just that warning.

deivid-rodriguez:
Yes, in the test I mean.

javierjulio (approved):
Thanks!

PR #713, opened 2022-01-19
Remove chandler

rafaelfranca:
It is not maintained anymore and now that GitHub has support to
create release notes automatically we don't need it.


deivid-rodriguez:
Hi, it's cool that GitHub now has built in support for this and I'm totally up for removing unmaintained gems and replacing them with more up to date alternatives. I just saw https://docs.github.com/en/repositories/releasing-projects-on-github/automatically-generated-release-notes, so we need to add some `.github/release.yml` file, correct?

deivid-rodriguez:
Or is https://github.com/activeadmin/inherited_resources/releases/tag/v1.13.1 what GitHub generates now by default? If that's the case, it's supercool!

rafaelfranca:
> Or is https://github.com/activeadmin/inherited_resources/releases/tag/v1.13.1 what GitHub generates now by default? If that's the case, it's supercool!

That is the default. You only need the release file if you want to automatic include/exclude something.

PR #714, opened 2022-01-23
Add Ruby 3.1 to CI matrix

alejandroperea:
This PR adds Ruby 3.1 to the CI matrix.

Rails version has been bumped to 7.0.1, that comes with `net-*` gems, not loaded by default in Ruby 3.1

Also was important to bump `thor` to its latest version to avoid `DidYouMean` deprecation warnings.


deivid-rodriguez (approved):
❤️

.github/mergify.yml
```diff
@@ -15,6 +15,8 @@ pull_request_rules:
       - check-success=test(ruby-3.0, rails_60)
       - check-success=test(ruby-3.0, rails_61)
       - check-success=test(ruby-3.0, rails_70)
+      - check-success=test(ruby-3.1, rails_61)
```
javierjulio, line 19:
```suggestion
      - check-success=test(ruby-3.1, rails_60)
      - check-success=test(ruby-3.1, rails_61)
```

javierjulio (approved):
Thanks! This is ready, just one update is needed.

.github/mergify.yml
alejandroperea: Good catch! I've missed that one. Added :)

javierjulio: Awesome thanks!

javierjulio (approved):
Thanks again! I'll merge.

PR #715, opened 2022-01-23
Drop support for Rails 5.2

alejandroperea:
I've seen in activeadmin repository that support for Rails 5.2 is being dropped, as it is very close to its EOL (~4-5 months) and only receives security patches.

This PR drops the support for that version and also includes Rails 6.1 and 7.0 in dependabot configuration.

If you think it is too early for this, we can close it and leave it for the future.


deivid-rodriguez (approved):
I'm totally up for this!

rafaelfranca (approved):

javierjulio (approved):
Thanks!

tagliala:
Hi,

https://github.com/activeadmin/inherited_resources/blob/19bb0b2f37c8b69d3a92ac1166f0601da52f3c23/lib/inherited_resources.rb#L28-L30

Is it possible that this check can be removed as described by the comment? I think that this could have been done after 5.1 removal

Ref: rails/rails#28402

deivid-rodriguez:
Happy to get such a change proposed @tagliala!

javierjulio:
Same here, thanks @tagliala!

PR #732, opened 2022-04-14
Drop support for Ruby 2.6

alejandroperea:
Ruby has served well but arrived to its EOL. Time to drop the support for this version!


deivid-rodriguez (approved):
Thanks @alejandroperea!

PR #738, opened 2022-04-26
Stop testing on JRuby

deivid-rodriguez:
JRuby is not yet compatible with Ruby 2.7.0, and that's the oldest MRI version that we officially support.


javierjulio (approved):
Thanks!

Issue #794, opened 2023-01-18
Need to support ruby 3.2


Damon1024:
Ruby 3.2.0 has been released for a while. It's now a good time to support ruby 3.2

PR #795, opened 2023-01-18
Allow using Ruby 3.2 #794

Damon1024:
Add build against Ruby 3.2
For issue https://github.com/activeadmin/inherited_resources/issues/794


javierjulio:
This has been updated, thanks!

PR #817, opened 2023-03-14
Opt-in for MFA requirement

tagliala:
Make the gem more secure by requiring that all privileged operations by any of the owners require OTP.

Ref: https://guides.rubygems.org/mfa-requirement-opt-in/


javierjulio (approved):
Thanks!

PR #858, opened 2023-08-22
Add frozen_string_literal magic comments

technicalpickles:
I spotted this while profiling my app with memory_profiler. InheritedResouces::UrlHelpers#define_params_helper showed up, where it allocated some 12320 blank strings (`''`).

It isn't much relatively by itself, but it adds up.

### DIFF
--- a/lib/inherited_resources/url_helpers.rb
+++ b/lib/inherited_resources/url_helpers.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
   # = URLHelpers
   #
--- a/lib/generators/rails/inherited_resources_controller_generator.rb
+++ b/lib/generators/rails/inherited_resources_controller_generator.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 require 'rails/generators/rails/scaffold_controller/scaffold_controller_generator'
 
 module Rails
--- a/lib/inherited_resources.rb
+++ b/lib/inherited_resources.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # This is here because responders don't require it.
 require 'rails/engine'
 require 'responders'
--- a/lib/inherited_resources/actions.rb
+++ b/lib/inherited_resources/actions.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
   # Holds all default actions for InheritedResouces.
   module Actions
--- a/lib/inherited_resources/base_helpers.rb
+++ b/lib/inherited_resources/base_helpers.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 # Whenever base is required load the dumb responder since it's used inside actions.
 require 'inherited_resources/blank_slate'
 
--- a/lib/inherited_resources/belongs_to_helpers.rb
+++ b/lib/inherited_resources/belongs_to_helpers.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
 
   # = belongs_to
--- a/lib/inherited_resources/blank_slate.rb
+++ b/lib/inherited_resources/blank_slate.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
   # An object from BlankSlate simply discards all messages sent to it.
   class BlankSlate
--- a/lib/inherited_resources/class_methods.rb
+++ b/lib/inherited_resources/class_methods.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
   module ClassMethods
 
--- a/lib/inherited_resources/dsl.rb
+++ b/lib/inherited_resources/dsl.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
   # Allows controllers to write actions using a class method DSL.
   #
--- a/lib/inherited_resources/engine.rb
+++ b/lib/inherited_resources/engine.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
   class Railtie < ::Rails::Engine
     config.inherited_resources = InheritedResources
--- a/lib/inherited_resources/polymorphic_helpers.rb
+++ b/lib/inherited_resources/polymorphic_helpers.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
 
   # = polymorphic associations
--- a/lib/inherited_resources/responder.rb
+++ b/lib/inherited_resources/responder.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
   class Responder < ActionController::Responder
     include Responders::FlashResponder
--- a/lib/inherited_resources/shallow_helpers.rb
+++ b/lib/inherited_resources/shallow_helpers.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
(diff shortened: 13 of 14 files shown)


lib/inherited_resources/url_helpers.rb
```diff
@@ -1,3 +1,6 @@
+# frozen_string_literal: true
+
+# # frozen_string_literal: true
```
javierjulio: Just a duplicate comment that can be removed.

```suggestion
# frozen_string_literal: true

```

lib/inherited_resources/version.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 module InheritedResources
-  VERSION = '1.13.1'.freeze
+  VERSION = '1.13.1'
```
javierjulio: This should remain as is because `frozen_string_literal` is opt-in even with Ruby 3.0 where it was no longer decided to make it the default.

```suggestion
  VERSION = '1.13.1'.freeze
```

tagliala: Hi, can I have a clarification?

Does it mean that even if the `# frozen_string_literal: true` magic comment is present in this file, there are chances that `VERSION` is not frozen?

javierjulio: Sure. Yes because you have to run Ruby with the option for enabling frozen string literals, otherwise the comment does nothing. It's opt-in. Originally, it was planned to make the option enabled by default in Ruby 3.0 but that was retracted. I don't know if they still intend to enable it by default in future major versions.

tagliala: So let's say that I have

### test-frozen.rb

```
# frozen_string_literal: true

puts 'hello'.frozen?
```

How I can run this file to get `false`?

I can see that there as option to opt-in and to allow a file without the magic comment to automatically freeze the strings, but I cannot see a "disable" option to override the magic comment

technicalpickles: To clarify a bit, `frozen_string_literal` has been around for awhile (initial googling suggests 2.3). It was proposed to turn that on by default for 3.0, as in, that is the behavior you get without the magic comment. But that didn't end up happening.

I haven't found a way that you can disable using `frozen_string_literal` for turning ON frozen strings, just enabling/disabling using frozen string literals globally without needing to use the magic comment.

javierjulio: @technicalpickles I'd like to get this out in a release if it helps but think we should do this update.

tagliala: @javierjulio I still do not get why is that needed. It also triggers a RuboCop offense:

```
$ cat version.rb 
# frozen_string_literal: true

module InheritedResources
  VERSION = '1.13.1'.freeze
end
```

```
$ rubocop version.rb 
Inspecting 1 file
C

Offenses:

version.rb:4:13: C: [Correctable] Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.
  VERSION = '1.13.1'.freeze
            ^^^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```

technicalpickles: I'm more interested in landing the rest of this PR than the specifics of this one file 😅  I made the change to use .freeze, but I agree with @tagliala .

I'm happy to do any other changes to get this accepted, but wanted to "Allow edits by maintianers" is checked if that is easier.

javierjulio: Then why not disable it in the file and leave the `.freeze` call in place? This would satisfy both sides. The point I was trying to make is that the change here assumes its only for those the are enabling the feature with the `--enable=frozen-string-literal` flag but if you are not, then this would be unfrozen now. Why not just disable the global setting in this single file using the magic comment and leave the `.freeze` call here? Am I missing something?

https://stackoverflow.com/a/37799399

technicalpickles: Sure, that works for me. I pushed up that change.

javierjulio (approved):
Thanks!

lib/inherited_resources/version.rb
tagliala: Hello,

sorry for the silence here

> The point I was trying to make is that the change here assumes its only for those the are enabling the feature with the --enable=frozen-string-literal flag but if you are not, then this would be unfrozen now

My interpretation for the above sentence is: "instead of adding a magic comment to each file, it is possible to frozen string literals with `--enable=frozen-string-literal` command line option".

At that point, the magic comment can work in the opposite way, also mentioned in the linked answer, by specifying `# frozen_string_literal: false`

So, I would say that if `# frozen_string_literal: true` is present, all literals present in the file are frozen, without the possibility for the developer to unfreeze them **(*)**

Moreover, there isn't a command line option `--disable=frozen-string-literal`, which would invalidate the magic comment, but even in those cases, if a user is deliberately adding that command line option, that should be honored so an unfrozen string would not be a problem

### Case 1: no `freeze`, no magic comment

```rb
ELSA = 'Elsa'
puts ELSA.frozen?
```

```sh
$ ruby elsa.rb 
false

$ ruby --enable=frozen-string-literal elsa.rb 
true
```

### Case 2: no `freeze`, magic comment with `true`

```rb
# frozen_string_literal: true

ELSA = 'Elsa'
puts ELSA.frozen?
```

```sh
$ ruby elsa_fsl_true.rb 
true

$ ruby --enable=frozen-string-literal elsa_fsl_true.rb 
true
```

### Case 3: no `freeze`, magic comment with `false`

```rb
# frozen_string_literal: false

ELSA = 'Elsa'
puts ELSA.frozen?
```

```sh
$ ruby elsa_fsl_false.rb 
false

$ ruby --enable=frozen-string-literal elsa_fsl_false.rb 
false
```

### Case 4: `freeze`, no magic comment

```rb
ELSA = 'Elsa'.freeze
puts ELSA.frozen?
```

```sh
$ ruby elsa_freeze.rb 
true

$ ruby --enable=frozen-string-literal elsa_freeze.rb 
true
```

### Case 5: `freeze`, magic comment `true`

Skipping example because it's trivial (all true)

### Case 6: `freeze`, magic comment `false`

```rb
# frozen_string_literal: false

ELSA = 'Elsa'.freeze
puts ELSA.frozen?
```

```sh
$ ruby elsa_freeze_fsl_false.rb 
true

$ ruby --enable=frozen-string-literal elsa_freeze_fsl_false.rb 
true
```

---

Since we are in **Case 2** (no `freeze`, magic comment `true`) it should be guaranteed that the string is frozen.

This should be the reason why RuboCop triggers an auto-correctable safe offense when developers try to add `freeze` on **Case 2** scenarios

```sh
$ cat elsa_fsl_true_freeze.rb 
# frozen_string_literal: true

ELSA = 'Elsa'.freeze
puts ELSA.frozen?
```

```sh
$ rubocop elsa_fsl_true_freeze.rb 
Inspecting 1 file
C

Offenses:

elsa_fsl_true_freeze.rb:3:8: C: [Correctable] Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.
ELSA = 'Elsa'.freeze
       ^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense autocorrectable
```

```sh
$ rubocop -a elsa_fsl_true_freeze.rb 
Inspecting 1 file
C

Offenses:

elsa_fsl_true_freeze.rb:3:8: C: [Corrected] Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.
ELSA = 'Elsa'.freeze
       ^^^^^^^^^^^^^

1 file inspected, 1 offense detected, 1 offense corrected
```

```sh
$ cat elsa_fsl_true_freeze.rb 
# frozen_string_literal: true

ELSA = 'Elsa'
puts ELSA.frozen?
```

---

**(*)** The only possibility for the developer to unfreeze would be to edit the file and remove the magic comment. If we fear such a possibility, then we should also `freeze` all other literals in all other files

PR #707, opened 2021-12-16
Allow Rails 7.0

tagliala:
Also:
- Tests against Rails 7.0 by default
- Updates all the Gemfile.lock

Supersedes #702:
- Does not remove Ruby 2.5 support
- Sets compatibility to `< 7.1` instead of `<= 7.0`

Hi, I'm aware of the other PR. I was working on this at the same time. On my fork, all tests are green: https://github.com/tagliala/inherited_resources/pull/2/checks 

Feel free to close, pick code from here, or move to #702

PR #873, opened 2023-10-05
Add Rails 7.1 support

tomascco:
I made this PR similar to #707, trying to keep the same structure:
- Updated the "main" Gemfile to Rails 7.1
- Moved the old Gemfile to the `gemfiles` folder
- Updated all Gemfiles

`rake test` passed on my machine with the same output (I think) before this PR

You may also feel free to modify, give feedbacks, etc.

Hope this helps 😄


javierjulio (approved):
Thanks! I'll do a separate update to remove the upper bound so we can avoid having to release, assuming no other changes are needed for support. That will close out #711.

javierjulio:
@tomascco there are several deprecation warnings as a result of updating to Rails 7.1. It would help to have a follow up to resolve those if you or someone else is able too. Thank you.

tomascco:
Sure, I'll try my best to tackle those warnings 😃

javierjulio:
Thank you, I appreciate it! If you have trouble feel free to post what you have in a PR and others can jump in.

PR #875, opened 2023-10-07
Fix test suite deprecations caused by Rails 7.1

tomascco:
In the following example:
```ruby
    Project.expects(:build).with({ 'these' => 'params' }).returns(mock_project(save: true))
    post :create, params: { student_id: '37', project: { these: 'params' } }
```
The `Project` instance was being mocked to expect `{ 'these' => 'params' }`, but when the code is being exercised, what comes is an `ActionController::Parameters` instance from the controller, so Mocha was comparing both to see if the expectation was being met, thus triggering the deprecated comparison between `ActionController::Parameters` and `Hash`.

This PR changes all problematic expectations to use `ActionController::Parameters` instead of hashes.

*There is another set of deprecations related to Mocha an positional arguments, but I think this PR is already pretty big, but if there is no problem, I can fix them here too.*

### DIFF
--- a/test/base_test.rb
+++ b/test/base_test.rb
@@ -65,6 +65,10 @@ module UserTestHelper
         user
       end
     end
+
+    def build_parameters(hash)
+      ActionController::Parameters.new(hash)
+    end
 end
 
 class IndexActionBaseTest < ActionController::TestCase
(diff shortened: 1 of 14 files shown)


test/base_test.rb
```diff
@@ -65,6 +65,10 @@ def method_missing(symbol, *arguments, &block)
         user
       end
     end
+
+    def build_ar_parameters(hash)
+      ActionController::Parameters.new(hash)
+    end
```
tomascco: This method is being repeated a lot, may we put it on `test_helper`, inside `ActiveSupport::TestCase`?

javierjulio: Sure, that's fine. I don't mind the repetition either. I'll leave that up to you.

Can you please rename the method to just `build_parameters`?

tomascco: Sure, and I left the tests self-contained, thanks for the feedback.

javierjulio (approved):
Thanks! I meant to mention about the Mocha warnings but I'm glad you didn't handle them. When I last reviewed and the issue in their repo, I believe these warnings are incorrect. A misunderstanding around the Ruby keyword arguments change in 3.0.

PR #884, opened 2023-10-25
Add RuboCop performance

tagliala:


gemfiles/rails_70/Gemfile.lock
```diff
@@ -219,6 +197,7 @@ GEM
 PLATFORMS
   arm64-darwin-22
   java
+  ruby
```
tagliala, line 200: Looks like an unintended change, but I can see ruby in arbre:

https://github.com/activeadmin/arbre/blob/181e9ebd4d4a12ae3db7a2f27dc4542d03586ef9/Gemfile.lock#L294

Let me know if this is not a problem

lib/inherited_resources/class_methods.rb
```diff
@@ -229,23 +229,23 @@ def belongs_to(*symbols, &block)
...
+        options[:polymorphic] = true
         belongs_to(*symbols, options, &block)
       end
 
       # A quick method to declare singleton belongs to.
       #
       def singleton_belongs_to(*symbols, &block)
         options = symbols.extract_options!
-        options.merge!(singleton: true)
+        options[:singleton] = true
```
tagliala, line 240: This line is already missing coverage

test/url_helpers_test.rb
```diff
@@ -11,6 +11,12 @@ class UniversesController < InheritedResources::Base
   defaults singleton: true, route_instance_name: 'universum'
 end
 
+class Galaxy < ModelBase
+end
+class GalaxiesController < InheritedResources::Base
+  singleton_belongs_to :universe
+end
+
```
tagliala: this is going to cover the missing line, even if there are not explicit tests, but I do not understand if this makes sense. Not familiar with "singleton belongs_to"

I guess this test needs a `test_url_helpers_on_singleton_belongs_to` but some guidance could help

gemfiles/rails_70/Gemfile.lock
javierjulio: I'm not sure what causes it. I'm sure it's fine as is. If it's a problem we can address the removal separately. Thanks!

test/url_helpers_test.rb
javierjulio: Neither am I. I haven't used it before. I know there was an issue submitted to ActiveAdmin regarding this feature. I'm fine with this as is if it meets the coverage requirements. Adding an actual test for the feature can occur separately, it shouldn't delay this. Let me know what you prefer to do here.

lib/inherited_resources/class_methods.rb
javierjulio: This has coverage now by adding a class for this feature but has no explicit test.

gemfiles/rails_70/Gemfile.lock
tagliala: This has been added because I have a setting to prefer ruby builds always on my bundler configuration

```
BUNDLE_FORCE_RUBY_PLATFORM: "true"
```

but should not be a problem

test/url_helpers_test.rb
```diff
@@ -106,7 +106,7 @@ class Spot < ModelBase
 end
 class SpotsController < InheritedResources::Base
   belongs_to :house do
-    belongs_to :dishwasher, singleton: true do
+    singleton_belongs_to :dishwasher do
```
tagliala, line 109: @javierjulio after a second look, I've noticed that singleton belongs to were actually tested but only through the `singleton: true` option.

Since `singleton: true` is used in multiple classes and tests, I've changed one to use the helper instead

javierjulio: @tagliala that's awesome! Sounds good to me. Thanks again for seeing this through. Let me know once this is ready for review.

javierjulio (approved):
Thanks!

PR #885, opened 2023-10-28
Add Rubocop Minitest

tagliala:
There are 37 `Minitest/NoAssertions` offenses left.

This is due to the fact that there are mocha expectations instead of "assert".

Example:

```rb
  def test_url_helpers_are_recreated_when_defaults_change
    BooksController.expects(:create_resources_url_helpers!).returns(true).once
    BooksController.send(:defaults, instance_name: 'string', collection_name: 'strings')
  end
```

No asserts here

The solution would be to:
- Disable this cop, but we may miss real offenses
- Add `assert true`

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -1,6 +1,7 @@
 ---
 
 require:
+  - rubocop-minitest
   - rubocop-performance
 
 AllCops:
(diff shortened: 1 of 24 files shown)


javierjulio:
@tagliala thanks! I think it would be best to leave the cop on for future changes. Would it be simple enough to add asserts for each of those warnings?

javierjulio:
@tagliala forgot that another option would be to just add disable comments for those type of tests which I'd also be fine with. I just think it would be best to leave the cop enabled.

tagliala:
I've disabled the cop when it is not required: 4eed462cb14117b9eb403e7abd7059abb163b45f

javierjulio (approved):
Thanks!

PR #902, opened 2024-03-03
Update Rubocop, enable new cops, fix offenses

tagliala:


lib/inherited_resources/base_helpers.rb
```diff
@@ -226,14 +226,14 @@ def get_resource_ivar #:nodoc:
...
-        instance_variable_set("@#{resource_instance_name}", resource)
+        instance_variable_set(:"@#{resource_instance_name}", resource)
       end
 
       # Get collection ivar based on the current resource controller.
       #
       def get_collection_ivar #:nodoc:
         if instance_variable_defined?(:"@#{resource_collection_name}")
-          instance_variable_get("@#{resource_collection_name}")
+          instance_variable_get(:"@#{resource_collection_name}")
```
javierjulio, line 236: This is the only line that has no test coverage. Do you think it would be possible to add a single test case for this?

tagliala: Done. It is not beautiful, but it does the job

javierjulio: No worries. What you did was exactly what I was thinking. This is great and just what we needed. Thank you.

javierjulio (approved):
Thanks!

PR #903, opened 2024-03-03
Support setting parent_controller

### LINKED ISSUE #618: Add an option to provide another superclass other than ApplicationController
`InheritedResource::Base` [inherits from](https://github.com/activeadmin/inherited_resources/blob/a863c1b58019f63e0ecf34960d4fe4737fc94bfb/app/controllers/inherited_resources/base.rb#L11) `::ApplicationController` and that's the source of a lot of issues for us during development. (We use it with ActiveAdmin). ApplicationController can be touched by many people and be reloaded but InheritedResource::Base is only loaded once. This leads to confusing behavior or breaking behavior with ActiveAdmin.

Also, we have an ApplicationController subclassing ActionController::API (for API applications), Inherited Resource/ActiveAdmin wouldn't work and we would _have to_ change our ApplicationController code for it to work.

See this related issue in ActiveAdmin repo: https://github.com/activeadmin/activeadmin/issues/2334

Request: Allowing InheritedResource::Base to inherit from a custom controller, or from ActionController::Base. 
For example: It can inherit from an `InheritedResource::BaseController`, and that file is generated. It can subclass ApplicationController but user can change it to subclass a different controller if they want.

jaynetics:
fixes #618

for the interface, i mirrored `Devise::parent_controller=`.

### DIFF
--- a/lib/inherited_resources.rb
+++ b/lib/inherited_resources.rb
@@ -26,6 +26,10 @@ module InheritedResources
   def self.flash_keys=(array)
     Responders::FlashResponder.flash_keys = array
   end
+
+  # Inherit from a different controller. This only has an effect if changed
+  # before InheritedResources::Base is loaded, e.g. in a rails initializer.
+  mattr_accessor(:parent_controller) { '::ApplicationController' }
 end
 
 ActiveSupport.on_load(:action_controller_base) do
--- a/test/parent_controller_test.rb
+++ b/test/parent_controller_test.rb
@@ -1,0 +1,21 @@
+require 'test_helper'
+
+def force_parent_controller(value)
+  InheritedResources.send(:remove_const, :Base)
+  InheritedResources.parent_controller = value
+  load File.join(__dir__, '..', 'app', 'controllers', 'inherited_resources', 'base.rb')
+end
+
+class ParentControllerTest < ActionController::TestCase
+  def test_setting_parent_controller
+    original_parent = InheritedResources::Base.superclass
+
+    assert_equal ApplicationController, original_parent
+
+    force_parent_controller('ActionController::Base')
+
+    assert_equal ActionController::Base, InheritedResources::Base.superclass
+  ensure
+    force_parent_controller(original_parent.to_s) # restore original parent
+  end
+end
--- a/test/parent_controller_test.rb
+++ b/test/parent_controller_test.rb
@@ -1,3 +1,4 @@
+# frozen_string_literal: true
 require 'test_helper'
 
 def force_parent_controller(value)
--- a/README.md
+++ b/README.md
@@ -168,6 +168,13 @@ class AccountsController < ApplicationController
 end
 ```
 
+By default, `InheritedResources::Base` will inherit from `::ApplicationController`. You can change this in a rails initializer:
+
+```ruby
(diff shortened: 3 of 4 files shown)


charnellh8 (approved):

jaynetics:
rebased to bring up-to-date with master.

@tagliala could we get this one merged?

test/parent_controller_test.rb
```diff
@@ -0,0 +1,19 @@
...
+
+class ParentControllerTest < ActionController::TestCase
+  def test_setting_parent_controller
+    original_parent = InheritedResources::Base.superclass
+
+    assert_equal ApplicationController, original_parent
+    force_parent_controller('ActionController::Base')
+
+    assert_equal ActionController::Base, InheritedResources::Base.superclass
+    force_parent_controller(original_parent.to_s) # restore original parent
```
tagliala, line 20: shouldn't this be in something like `ensure` ?

jaynetics: @tagliala done.

lib/inherited_resources.rb
```diff
@@ -25,6 +25,10 @@ module InheritedResources
   def self.flash_keys=(array)
     Responders::FlashResponder.flash_keys = array
   end
+
+  # Inherit from a different controller. This only has an effect if changed
+  # before InheritedResources::Base is loaded, e.g. in a rails initializer.
+  mattr_accessor(:parent_controller) { '::ApplicationController' }
```
tagliala, line 32: can I ask if this is the same as 

https://github.com/heartcombo/devise/blob/a259ff3c28912a27329727f4a3c2623d3f5cb6f2/lib/devise.rb#L244-L245

?

Ref: https://github.com/rails/rails/blob/56fbc632a378acc40d99f5b4be321f80a5d8b8c4/activesupport/lib/active_support/core_ext/module/attribute_accessors.rb#L208-L212

jaynetics: @tagliala i'd say its equivalent – the block is evaluated and the class variable is set immediately: 

https://github.com/rails/rails/blob/56fbc632a378acc40d99f5b4be321f80a5d8b8c4/activesupport/lib/active_support/core_ext/module/attribute_accessors.rb#L69-L70

i think the devise code is just like that because when José Valim wrote these lines 13 years ago, mattr_accessor did not yet support passing a default value:

https://github.com/rails/rails/blob/bf526c2dbeb73bf11553004e43889a804b72866d/activesupport/lib/active_support/core_ext/module/attribute_accessors.rb#L60

tagliala:
I think this may be an interesting change, I may need this for a use case where I want to alias `current_admin_user` to `current_user` and add some other admin-specific actions

jaynetics:
My use case is the same as in the original issue: running rails in API mode. 

I've been using this fork and the setting `parent_controller = 'ActionController::Base'` in production for 8 months now without issues.

test/parent_controller_test.rb
```diff
@@ -0,0 +1,21 @@
+require 'test_helper'
```
javierjulio, line 2:
```suggestion
# frozen_string_literal: true
require 'test_helper'
```

javierjulio (approved):
Thanks!

PR #909, opened 2024-04-01
Test against Ruby 3.3

tagliala:


javierjulio (commented):
@tagliala thanks for adding this in. Looks like our Nokogiri version is behind a minor version and the current one supports Ruby 3.3 so a simple update should fix the build.

tagliala:
Sure, I'll do that later or tomorrow on my dev machine

javierjulio (approved):
Thanks!

PR #916, opened 2024-06-01
Make lockfile platforms easier to maintain

tagliala:
This should work equally well and has less churn when maintainers upgrade their OS or switch platforms.

Recent versions of Bundler will "optimize" this, and as long as no variants specific to the macOS version exist, it will lock only arch + OS, but not the specific macOS version.

These should not get added in the future again as long as we use a recent enough version of Bundler, and for this reason locked Bundler has been bumped to 2.5.11

Ref: activeadmin/activeadmin#8344


tagliala:
I would like to add credits in the commit to @deivid-rodriguez, but I don't know how to do that

javierjulio:
@tagliala no worries, he won't mind. He'll just be happy that this is making its way to other projects too. Thank you for doing this. I appreciate it.

javierjulio (approved):
Thanks!

deivid-rodriguez:
@tagliala Indeed I don't mind at all, thanks for adding this actually :)!

For next time, the way I usually do this is to find user name and email of the user I want to credit through `git log`, then I amend the commit including authorship. See https://docs.github.com/en/pull-requests/committing-changes-to-your-project/creating-and-editing-commits/creating-a-commit-with-multiple-authors.

PR #863, opened 2023-08-31
update responders config to match default error_status & redirect_status codes for current rails version

a-nickol:
Update responder class to the current defaults for responder gem. 

Fixes #862

PR #918, opened 2024-06-01
Change return codes to new Responders defaults

### LINKED ISSUE #862: `Create` & `Update` should return `status code != 200` for failed requests
To make `inherited_resources` work with the current version of rails out of the box, it is not allowed to not redirect if status code is 200 on a submitted form.
 
Since on a failure we want to render either `:new` or `:edit`, we are not allowed send status code 200, since this will result in the following error:

```
Error: Form responses must redirect to another location
    at _FormSubmission.requestSucceededWithResponse (application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1298:24)
    at FetchRequest.receive (application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1068:23)
    at FetchRequest.perform (application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1046:27)
formSubmissionErrored @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:2854
requestSucceededWithResponse @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1299
receive @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1068
perform @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1046
await in perform (async)
start @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1258
submitForm @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:2795
formSubmitted @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:3504
FormSubmitObserver.submitBubbled @ application-c20cb2d408231ec629db20f6bfe840b8b84f28b26acee36de46dd6b684b8aa43.js:1457
```

**Workaround:**

```
  def create
    create! do |success, failure|
      failure.html { render(:new, status: :unprocessable_entity) }
    end
  end

  def update
    update! do |success, failure|
      failure.html { render(:edit, status: :unprocessable_entity) }
    end
  end
```

tagliala:
Change default error and redirect status codes
Update responder class to the current defaults for responder gem

Close: https://github.com/activeadmin/inherited_resources/issues/862

Ref: https://github.com/heartcombo/responders/pull/240

Supersedes #863

---

I'm doing some tests with this branch and `$ bin/rake local server` with Active Admin main

|Action|Method|Status Before|Status After|Match Rails|
|:-|:-|:-|:-|:-|
|Create category|POST|302 Found|303 See Other|❌ (302, but 303 is better)|
|Create category (fail validations)|POST|200 OK|422 Unprocessable Entity|✅|
|Update category|POST|302 Found|303 See Other|❌ (302, but 303 is better)|
|Update category (fail validations)|POST|200 OK|422 Unprocessable Entity|✅|
|Destroy category|POST|302 Found|303 See Other|✅|

Rails by default returns `302` on successful create and update, and `303` only on destroy. However, according to HTTP/1.1 standard, it appears that `303` is the correct status code to redirect from a `POST/PATCH/PUT/DELETE` to a `GET` and clients do treat `302` like a `303`, by using `GET` instead of using the original method

### References that motivate the decision:

- rails/rails#45393
- hotwired/turbo#84
- https://github.com/heartcombo/responders?tab=readme-ov-file#configuring-error-and-redirect-statuses
  > Note: the application responder generated for new apps already configures a different set of defaults: 422 Unprocessable Entity for errors, and 303 See Other for redirects. Responders may change the defaults to match these in a future major release.
- https://en.wikipedia.org/wiki/HTTP_302
  > **Many web browsers implemented this code in a manner that violated this standard**, changing the request type of the new request to [GET](https://en.wikipedia.org/wiki/HTTP_GET_request), regardless of the type employed in the original request (e.g. [POST](https://en.wikipedia.org/wiki/POST_(HTTP))).[[1]](https://en.wikipedia.org/wiki/HTTP_302#cite_note-1) For this reason, **HTTP/1.1 (RFC [2616](https://datatracker.ietf.org/doc/html/rfc2616)) added the new status codes [303](https://en.wikipedia.org/wiki/HTTP_303) and [307](https://en.wikipedia.org/wiki/HTTP_307) to disambiguate between the two behaviours, with 303 mandating the change of request type to GET**, and 307 preserving the request type as originally sent. Despite the greater clarity provided by this disambiguation, the 302 code is still employed in web frameworks to preserve compatibility with browsers that do not implement the HTTP/1.1 specification.[[2]](https://en.wikipedia.org/wiki/HTTP_302#cite_note-2)

  (**emphasis** mine)

- https://datatracker.ietf.org/doc/html/rfc2616#section-10.3.4
  > If the 302 status code is received in response to a request other
   than GET or HEAD, the user agent MUST NOT automatically redirect the
   request unless it can be confirmed by the user, since this might
   change the conditions under which the request was issued.
  > 
  > Note: [RFC 1945](https://datatracker.ietf.org/doc/html/rfc1945) and [RFC 2068](https://datatracker.ietf.org/doc/html/rfc2068) specify that the client is not allowed
      to change the method on the redirected request.  However, **most
      existing user agent implementations treat 302 as if it were a 303
      response, performing a GET on the Location field-value regardless
      of the original request method**. The status codes 303 and 307 have
      been added for servers that wish to make unambiguously clear which
      kind of reaction is expected of the client.

  (**emphasis** mine)

Default responder created by `rails g responders:install`: https://github.com/heartcombo/responders/blob/9bdc60dfbfa8001641c1c4df7bc73c3fc2a4cf41/lib/generators/responders/install_generator.rb#L10-L25

### DIFF
--- a/lib/inherited_resources/responder.rb
+++ b/lib/inherited_resources/responder.rb
@@ -3,5 +3,8 @@
 module InheritedResources
   class Responder < ActionController::Responder
     include Responders::FlashResponder
+
+    self.error_status = :unprocessable_entity
+    self.redirect_status = :see_other
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 * Test against Rails 7.2 [#921][].
+* Change return codes to new Responders defaults [#918][].
 
 ## Version 1.14.0
 
@@ -229,4 +230,5 @@ _No changes_.
 * First release. Support to I18n, singleton controllers, polymorphic controllers, belongs_to, nested_belongs_to and url helpers.
 
 [#873]: https://github.com/activeadmin/inherited_resources/pull/873
+[#918]: https://github.com/activeadmin/inherited_resources/pull/918
 [#921]: https://github.com/activeadmin/inherited_resources/pull/921
--- a/test/aliases_test.rb
+++ b/test/aliases_test.rb
@@ -107,7 +107,7 @@ class AliasesTest < ActionController::TestCase
     @controller.stubs(:resource_url).returns('http://test.host/')
     post :create
 
-    assert_response :success
+    assert_response :unprocessable_entity
     assert_equal "New HTML", @response.body.strip
   end
 
@@ -117,7 +117,7 @@ class AliasesTest < ActionController::TestCase
     @controller.stubs(:resource_url).returns('http://test.host/')
     post :create
 
-    assert_response :success
+    assert_response :unprocessable_entity
     assert_equal "New HTML", @response.body.strip
   end
 
--- a/test/base_test.rb
+++ b/test/base_test.rb
@@ -246,7 +246,7 @@ class CreateActionBaseTest < ActionController::TestCase
     User.stubs(:new).returns(mock_user(save: false, errors: {some: :error}))
     post :create
 
-    assert_response :success
+    assert_response :unprocessable_entity
     assert_equal "New HTML", @response.body.strip
   end
 
@@ -322,7 +322,7 @@ class UpdateActionBaseTest < ActionController::TestCase
     User.stubs(:find).returns(mock_user(update: false, errors: {some: :error}))
     put :update, params: { id: '42' }
 
-    assert_response :success
+    assert_response :unprocessable_entity
     assert_equal "Edit HTML", @response.body.strip
   end
 
@@ -370,6 +370,7 @@ class DestroyActionBaseTest < ActionController::TestCase
     User.stubs(:find).returns(mock_user(destroy: false, errors: { fail: true }))
     delete :destroy, params: { id: '42' }, format: :js
 
+    assert_response :unprocessable_entity
     assert_equal 'User could not be destroyed.', flash[:alert]
   end
 
@@ -378,6 +379,7 @@ class DestroyActionBaseTest < ActionController::TestCase
     @controller.expects(:collection_url).returns('http://test.host/')
     delete :destroy, params: { id: '42' }
 
+    assert_response :see_other
     assert_redirected_to 'http://test.host/'
   end
 
@@ -386,6 +388,7 @@ class DestroyActionBaseTest < ActionController::TestCase
     @controller.expects(:collection_url).returns('http://test.host/')
     delete :destroy, params: { id: '42' }
 
+    assert_response :see_other
     assert_redirected_to 'http://test.host/'
   end
 end
--- a/test/customized_base_test.rb
+++ b/test/customized_base_test.rb
@@ -141,7 +141,7 @@ class CreateActionCustomizedBaseTest < ActionController::TestCase
     Car.stubs(:create_new).returns(mock_car(save_successfully: false, errors: {some: :error}))
     post :create
 
-    assert_response :success
+    assert_response :unprocessable_entity
     assert_equal "New HTML", @response.body.strip
   end
 end
@@ -169,7 +169,7 @@ class UpdateActionCustomizedBaseTest < ActionController::TestCase
     Car.stubs(:get).returns(mock_car(update_successfully: false, errors: {some: :error}))
     put :update, params: { id: '42' }
 
-    assert_response :success
+    assert_response :unprocessable_entity
     assert_equal "Edit HTML", @response.body.strip
   end
 end


tagliala:
@javierjulio 

As per https://github.com/heartcombo/responders?tab=readme-ov-file#configuring-error-and-redirect-statuses, newly generated application responders default to 

```rb
self.error_status = :unprocessable_entity
self.redirect_status = :see_other
```

(and also include `include Responders::HttpCacheResponder`)

So it may have sense to change those fields, regardless of Turbo support, and release a major version

lib/inherited_resources/responder.rb
```diff
@@ -3,5 +3,10 @@
 module InheritedResources
   class Responder < ActionController::Responder
     include Responders::FlashResponder
+    include Responders::HttpCacheResponder
```
javierjulio: This doesn't appear to be a new default but [was always part of the generator for over a decade](https://github.com/heartcombo/responders/commit/4d094297022f6ab6db6d4a9973dc7453133c1f1c#diff-544fda81eeab7956b30883ad942d73dab5ba15b53c593f19dfd6e892680e85b4).

Based on [the class documentation details](https://github.com/heartcombo/responders/blob/main/lib/responders/http_cache_responder.rb#L4-L12) and also [what the code does internally](https://github.com/heartcombo/responders/blob/main/lib/responders/http_cache_responder.rb#L27-L29), I do not want to introduce this as a default here. Would be safest to just focus on only the Turbo related changes.

```suggestion
```

javierjulio (commented):
Can you please retest this without the `Responders::HttpCacheResponder` inclusion? When retesting, can you please make sure that forms for both new/create and edit/update, work as expected? Both success and failure scenarios. I just want to make sure there isn't any regressions there for ActiveAdmin. I don't believe there would be.

lib/inherited_resources/responder.rb
tagliala: I do agree. I've asked to remove this in the other PR, but then I've seen it was a default and reintroduce it.

I will add a note in the commit that we are deliberately not including that line

tagliala:
Updated with a ref to a9fb8e1cbf346cf217483b0a45ce04384b10c41e

javierjulio (commented):
Thank you for including the reference. It makes sense why it was changed. Referencing the "why" is always helpful so I appreciate it.

How did retesting this go with ActiveAdmin? I'd like to ensure that submitting forms, both successful and with validation errors, works as expected. Essentially, no regressions.

Do you think we should release this in a major version? ActiveAdmin is locked to v1.x for InheritedResources so earlier version won't be able to use it so releasing a major version of this is safe. We could permit it in ActiveAdmin v4 though since it's still beta.

tagliala:
> How did retesting this go with ActiveAdmin? I'd like to ensure that submitting forms, both successful and with validation errors, works as expected. Essentially, no regressions.

According to my tests in the default test app, there are no visible changes. I will run active admin specs against this branch

> Do you think we should release this in a major version?

It is a potential breaking change with some custom javascript expecting a 200 from the server, so in any case it makes sense to release in a major version

> ActiveAdmin is locked to v1.x for InheritedResources so earlier version won't be able to use it so releasing a major version of this is safe. We could permit it in ActiveAdmin v4 though since it's still beta.

Probably it is a good choice

javierjulio (approved):
Thanks! Let's release as v2 then.

tagliala:
All Active Admin specs (both `master` and `3-0-stable`) pass against this branch

---

## Screenshots

### Before

image

### After

image

tagliala:
@javierjulio rebased 

- Should I bump the major in this branch?
- Should we release a beta with this change?

javierjulio:
@tagliala awesome thanks. I think this is safe as a major release, no beta release needed. You are welcome to bump the major version here or in another PR, I'm fine with either. Thank you.

PR #921, opened 2024-06-10
Test against Rails 7.2

tagliala:


.github/workflows/ci.yaml
```diff
@@ -8,48 +8,48 @@ on:
...
-          - name: ruby-3.1
-            value: 3.1
-          - name: ruby-3.2
-            value: 3.2
-          - name: ruby-3.3
-            value: 3.3
+          - "3.0"
+          - "3.1"
+          - "3.2"
+          - "3.3"
```
tagliala: sorry, I've removed this extra complexity because it was difficult to `exclude`ing 3.0/rails_72 and action linter was complaining about a hash value in `ruby`

javierjulio (approved):
Thanks!

PR #929, opened 2024-07-28
Use https where possible and fix redirects

tagliala:
Issues identified with the help of `awesome_bot` gem


javierjulio (approved):
Thanks!

PR #933, opened 2024-08-26
Add frozen string literal magic comment (potential breaking change)

tagliala:
This cop was disabled probably by mistake, since it is present in both
Arbre and Active Admin gems

The only production code affected is:
- app/controllers/inherited_resources/base.rb
- lib/inherited_resources/version.rb

Change to `version.rb` is safe.

After inspecting the source code of `InheritedResources` module, the
change seems safe, but it can represent a breaking change in third-party application.

Since we are bumping the major version number, it would be fine to break
if documented.

Fixed with `rubocop -A`

Also adds `Style/Encoding` cop for parity with Arbre

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -373,5 +373,13 @@ Performance/UnfreezeString:
 Performance/UriDefaultParser:
   Enabled: true
 
+Style/Encoding:
+  Enabled: true
+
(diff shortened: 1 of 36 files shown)


javierjulio (approved):
Thanks!

PR #939, opened 2024-09-19
Prefer require_relative for internal requires

tagliala:
`require_relative` is preferred over `require` for files within the same project because it uses paths relative to the current file, making code more portable and less dependent on the load path.

This change updates internal requires to use `require_relative` for consistency, performance, and improved portability.

Ref:
- activeadmin/arbre#622


deivid-rodriguez (approved):
❤️

PR #941, opened 2024-09-26
Refactor polymorphic keys method

tagliala:
Use `select` instead of `map.compact` to:
- Eliminate unnecessary object creation and array compaction
- Maintain identical functionality
- Improve readability and performance

```
Comparison (IPS):
                keys:  1716164.8 i/s
            keys_old:  1531182.7 i/s - 1.12x  (± 0.00) slower

Comparison (Memory):
                keys:        120 allocated
            keys_old:        160 allocated - 1.33x more
```


javierjulio (approved):
Thanks!

PR #946, opened 2024-11-09
Drop Rails < 7.0 support

tagliala:
Additionally, update the changelog by removing all trailing punctuation.
Fix the test to ensure it works as described and is consistent with
Arbre and ActiveAdmin.


inherited_resources.gemspec
```diff
@@ -25,7 +25,7 @@ Gem::Specification.new do |s|
   s.required_ruby_version = '>= 3.1'
 
   s.add_dependency("responders", ">= 2")
-  s.add_dependency("actionpack", ">= 6.0")
-  s.add_dependency("railties", ">= 6.0")
+  s.add_dependency("actionpack", ">= 7.0")
+  s.add_dependency("railties", ">= 7.0")
```
tagliala, line 29: Apparently 6.0 was still supported, but not tested

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## Unreleased
 
+* Remove support for Ruby `< 3.1`.
+* Remove support for Rails `< 7.0`.
+* Test against Rails 8.0.
+
```
tagliala: Pretty unexpected that this is not failing like Arbre

```
Failures:

  1) Changelog entry does not end with a punctuation
     Failure/Error: expect(entry).not_to match(/\.$/)

       expected "* Remove support for Ruby `< 3.1`.\n" not to match /\.$/
       Diff:
       @@ -1 +1 @@
       -/\.$/
       +* Remove support for Ruby `< 3.1`.
     # ./spec/changelog_spec.rb:24:in `block (4 levels) in <top (required)>'
     # ./spec/changelog_spec.rb:[23](https://github.com/activeadmin/arbre/actions/runs/11757484591/job/32754790006?pr=639#step:6:24):in `each'
     # ./spec/changelog_spec.rb:23:in `block (3 levels) in <top (required)>'
```

https://github.com/activeadmin/arbre/actions/runs/11757484591/job/32754790006?pr=639

tagliala: Ok, the spec was wrong here. I've fixed the intent. There are a lot of changes in the changelog, but now it is uniform with Arbre and ActiveAdmin (no punctuations at the end of the line). If you prefer, I can invert the spec description and leave dots (but that would not allow easy copy & paste from changelogs, like I did)

javierjulio (approved):
Thanks!

CHANGELOG.md
javierjulio: No, I think it’s fine as you have it as it would be easiest for us. Thank you. If anything I may just remove the spec in the future.

PR #950, opened 2024-12-29
Test against Ruby 3.4

tagliala:
Additionally: return implicit `nil` from `SingletonHelpers#collection`
to keep the same coverage results as 3.3


tagliala:
I'm getting different code coverage between 3.3.6 and 3.4.1

```
# 3.3.6
Coverage report generated for Unit Tests to ./coverage/coverage.xml. 552 / 558 LOC (98.92%) covered

# 3.4.1
Coverage report generated for Unit Tests to ./coverage/coverage.xml. 559 / 566 LOC (98.76%) covered
```

3.4.1 detects:
- 3 more lines detected in `lib/inherited_resources/base_helpers.rb` (112 vs 109)
- 2 more lines detected in `lib/inherited_resources/singleton_helpers.rb` (12 vs 10)

In singleton_helpers, `def collection; nil; end` does not seem to be covered in 3.4.1, and it looks to be correct, because in 3.3.6 

```diff
       def collection
-        nil
+        raise
       end
```

does not raise. So 3.3.6 was reporting wrong coverage

javierjulio:
@tagliala thank you. Can you please try removing the `nil` line so the method just becomes an empty method? It would have the same return value and may appease Codecov here without need for further changes.

javierjulio (approved):
Thanks!

PR #952, opened 2025-01-12
Update dependencies

tagliala:
Additionally, enable a new cop to remove the following warning:

```
The following cops were added to RuboCop, but are not configured. Please set Enabled to either `true` or `false` in your `.rubocop.yml` file.
```


javierjulio (approved):
Thanks!

PR #959, opened 2025-07-03
Update dependencies and improve Dependabot config

tagliala:
- Remove Rails 6.1, which is not supported anymore
- Add new Rails versions
- Sort dependencies alphabetically

Similar to activeadmin/arbre#683


javierjulio (approved):
Thanks!

PR #1006, opened 2026-07-19
Update SimpleCov to 1.0

tagliala:
Remove the explicit filter on `/test`, which is now ignored by default.


javierjulio (approved):
Thanks!

